PR #6, opened 2026-04-22
feat(qed): ratchet — upgrade-safety + mainnet-readiness lint

saicharanpogul:
## Summary

Following up on our DM — this wires [ratchet](https://github.com/saicharanpogul/ratchet) into the qedgen pipeline as the post-codegen step you asked for.

Proofs prove correctness. Ratchet proves **deployability**.

The `#[qed(verified)]` hash-stamp catches when a function *body* drifts from its proof. It doesn't catch when an `#[account]` struct is renamed — the proof stays green, but the Anchor discriminator is now different and every existing account on-chain is orphaned. Same family as `shape_only_cpi` and `qedgen reconcile`: "X looks done, but Y gap remains." Ratchet fills the third seat.

Dogfooded on `examples/rust/escrow` before writing any code — flagged three real future-upgrade landmines in a program that's already formally verified:

```
UNSAFE    P001  account-missing-version-field        account:EscrowState
UNSAFE    P002  account-missing-reserved-padding     account:EscrowState
additive  P003  account-missing-discriminator-pin    account:EscrowState/discriminator
verdict: UNSAFE — `escrow` has future-upgrade concerns
```

### `qedgen readiness --idl <anchor.json>`
Pre-deploy preflight on one Anchor IDL. Runs the six P-rules — missing `version: u8` prefix, no `_reserved` padding, unpinned discriminators (account + event), name collisions, writable accounts with no signer. Human or `--json` output. Exit codes `0 = additive/safe, 1 = breaking, 2 = unsafe`.

### `qedgen check-upgrade --old <old.json> --new <new.json>`
Diff mode — R001–R016. Catches field reorders, discriminator changes, orphaned accounts, PDA seed drift, signer/writable tightening. Supports `--unsafe <flag>` (repeatable), `--migrated-account <Name>`, `--realloc-account <Name>` for acknowledging intentional unsafe changes. Same human / JSON / exit-code surface as `readiness`.

### CI integration: `--ci-ratchet`
```bash
qedgen codegen --spec my_program.qedspec --ci --ci-ratchet target/idl/my_program.json
```
Adds a `{{RATCHET_STEP}}` placeholder to `templates/verify.yml` that expands to:
```yaml
      - name: Ratchet readiness lint
        run: qedgen readiness --idl target/idl/<program>.json
```
Empty-string replacement when `--ci-ratchet` is omitted, so existing codegen output is unchanged for users who don't opt in. The escrow example's committed `verify.yml` snapshot is regenerated with the flag to demonstrate.

### Architecture note — library, not subprocess
`qedgen` embeds `solana-ratchet-core` and `solana-ratchet-anchor` (no-default-features, so `ureq` drops out because we read IDLs from disk, never from RPC). Single binary for end-users; ratchet version pinned via `Cargo.lock`; no PATH / install-step surprises.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace` (183 tests total, 6 new in `ratchet.rs` covering readiness + check-upgrade + exit-code mapping + IO/parse error paths)
- [x] `bash scripts/check-readme-drift.sh` — 15 / 15 subcommands documented
- [x] End-to-end: `qedgen readiness --idl target/idl/escrow.json` on the committed escrow example — findings match the pre-PR dogfood run above
- [x] End-to-end: `qedgen codegen --ci --ci-ratchet …` regenerates `examples/rust/escrow/.github/workflows/verify.yml` cleanly; committed snapshot in this PR

## Files touched

| File | Change |
|---|---|
| `crates/qedgen/Cargo.toml` | + `ratchet-core = "0.3"`, `ratchet-anchor = "0.3"` (no-default-features, aliased via `package =` so the import path stays `ratchet_core::` / `ratchet_anchor::`) |
| `crates/qedgen/src/ratchet.rs` | New module — `run_readiness`, `run_check_upgrade`, `exit_code`, human + JSON printers, 6 inline `#[test]`s (250 LOC) |
| `crates/qedgen/src/main.rs` | `mod ratchet;`, `Readiness { idl, json }` + `CheckUpgrade { old, new, unsafes, migrated_accounts, realloc_accounts, json }` variants on `Commands`, two new match arms next to `Verify`, `--ci-ratchet` flag on `Codegen` + template emission |
| `templates/verify.yml` | `{{RATCHET_STEP}}` placeholder appended after the existing "Check spec coverage" step |
| `examples/rust/escrow/.github/workflows/verify.yml` | Regenerated with `--ci-ratchet target/idl/escrow.json` |
| `README.md` | New row in "What it verifies" table + "Deploy-safety lint (ratchet)" section under Usage + `--ci-ratchet` example under "Generate CI workflow" |

**Total: 250 LOC added, 7 files, one themed bundle — matches the repo's existing PR cadence.**

## Out of scope (follow-ups if you want them)

- Making `--ci-ratchet` default-on when `--all` is set. Deliberately opt-in for this PR so the rollout is non-breaking.
- A `qedgen reconcile`-style integration that cross-references `#[qed(verified)]` hashes with ratchet's R006 findings (rename detection at the attribute level). Interesting but scope creep.
- Publishing `solana-ratchet-wasm` so the skill could also run inside the browser-based qedgen UI — ratchet already has a WASM build, just isn't on crates.io yet.

## Why now

Readiness mode landed in ratchet v0.3.0 last week, which is what makes the pre-deploy story actually work (before v0.3.0, ratchet was diff-only, which isn't useful until you've already deployed). Happy to amend this PR in any direction — scope, naming, commit splits — whatever matches your review flow best.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bs58"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf88ba1141d185c399bee5288d850d63b8369520c1eafc32a0430b5b6c287bf4"
+dependencies = [
+ "tinyvec",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -238,6 +247,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "curve25519-dalek"
+version = "4.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97fb8b7c4503de7d6ae7b42ab72a5a59857b4c937ec27a3d4539dba95b5ab2be"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "curve25519-dalek-derive",
+ "fiat-crypto",
+ "rustc_version",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
+name = "curve25519-dalek-derive"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -279,6 +314,12 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
+
+[[package]]
+name = "fiat-crypto"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28dea519a9695b9977216879a3ebfddf92f1c08c05d984f8996aecd6ecdc811d"
 
 [[package]]
 name = "filetime"
@@ -873,6 +914,8 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "solana-ratchet-anchor",
+ "solana-ratchet-core",
  "syn",
  "tar",
  "tempfile",
@@ -931,7 +974,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1103,6 +1146,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "357703d41365b4b27c590e3ed91eabb1b663f07c4c084095e60cbed4362dff0d"
 
 [[package]]
+name = "rustc_version"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfcb3a22ef46e85b45de6ee7e79d063319ebb6594faafcf1c225ea92ab6e9b92"
+dependencies = [
+ "semver",
+]
+
+[[package]]
 name = "rustix"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1167,6 +1219,12 @@ name = "scopeguard"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
+
+[[package]]
+name = "semver"
+version = "1.0.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a7852d02fc848982e0c167ef163aaff9cd91dc640ba85e263cb1ce46fae51cd"
 
 [[package]]
 name = "serde"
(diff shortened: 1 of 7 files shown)


abishekk92 (approved):
Thanks for the thorough writeup and for dogfooding on `examples/rust/escrow` before landing this — the pre-PR finding list is exactly the kind of signal that makes the framing ("proofs prove correctness; ratchet proves deployability") land. Library embedding over subprocess is the right call, the `default-features = false` drop of `ureq` is thoughtful, and the test coverage (happy path + severity mapping + IO/parse errors + R007 diff) is solid.

Approving. Two small asks before merge — both minor, neither blocking in spirit:

1. **Exit code 3 is documented but never emitted.** The module comment in `crates/qedgen/src/ratchet.rs` claims `3 = qedgen-level error before the engine ran (caller-side)`, but every error path is an `anyhow::Error` that bubbles to `main` and exits 1 via clap's default. Either wire it (`if let Err(e) = … { exit(3) }` in the two match arms in `main.rs`) or drop the `3 = …` line from the comment — pick whichever you prefer, just want the doc and behavior to agree.

2. **Template expansion isn't unit-tested.** A ~10-line snapshot test asserting that `{{RATCHET_STEP}}` round-trips correctly for both set and unset `--ci-ratchet` would be cheap insurance — it's easy for a future `templates/verify.yml` refactor to silently drop or mangle the placeholder, and the existing tests don't catch that.

A few other things I noticed but don't need to hold the merge:

- `print_human` → stderr and `print_json` → stdout in `ratchet.rs` is defensible (matches ratchet's own CLI) but unusual enough inside one subcommand to warrant a one-line comment noting the intent.
- Committed `examples/rust/escrow/.github/workflows/verify.yml` has a trailing blank line from the `{{RATCHET_STEP}}` substitution when unset. Minor cosmetic; fixable by having the "on" substitution absorb the leading newline.
- README references `ratchet list-rules`, but qedgen embeds ratchet as a library so users won't have the CLI on PATH. Worth either linking upstream docs or adding a `qedgen readiness --list-rules` passthrough in a follow-up PR.
- `curve25519-dalek` + `fiat-crypto` are new transitive deps; worth a glance (separately) at whether more `default-features` can be peeled off `ratchet-anchor`.
- Flagging for your awareness: ratchet is Anchor-IDL-only, and the README's "Deploy safety" row reads as if it applies across the board. One qualifying clause ("Anchor programs") would sharpen it.

Happy to merge once (1) and (2) are in. Thanks again for the contribution — this is a clean, well-scoped addition and the PR body itself is one of the better ones I've reviewed recently.

abishekk92 (approved):

saicharanpogul:
Thanks — both asks addressed in `926f34e`:

**1. Exit code 3 wired.** The `Readiness` and `CheckUpgrade` match arms in `main.rs` now catch the `anyhow::Error` from `run_readiness` / `run_check_upgrade` and `std::process::exit(3)` explicitly. Rule-engine findings still map to `0 / 1 / 2` via `ratchet::exit_code`. Verified on the happy path and with a missing-IDL path:

```
$ qedgen readiness --idl /does/not/exist.json; echo exit=$?
Error: reading /does/not/exist.json: No such file or directory (os error 2)
exit=3

$ qedgen readiness --idl /tmp/ratchet-dogfood/escrow.json > /dev/null 2>&1; echo exit=$?
exit=2
```

**2. Template expansion tested.** Factored the substitution into `expand_ci_template(template, verify_step, ratchet_step)` and added three snapshot tests in `crates/qedgen/src/main.rs`:

- `ci_template_unset_placeholders_produce_clean_workflow` — no `{{…}}` residue, no optional steps, EOF is a single `\n`.
- `ci_template_ratchet_step_injects_readiness_job` — `{{RATCHET_STEP}}` round-trips correctly when set.
- `ci_template_both_steps_coexist_without_collision` — sBPF step precedes `Build proofs`, ratchet step follows `Check spec coverage`.

Also folded in the three non-blocking items while I was in there:

- Added a comment to `ratchet::print_human` / `print_json` noting the stderr-vs-stdout split and the agent-redirect rationale.
- `expand_ci_template` now normalises trailing whitespace, so the regenerated escrow `verify.yml` no longer ships a stray blank line at EOF (one trailing `\n`, verified via `xxd`).
- Qualified the "Deploy safety" README row as *"On-chain shape for Anchor programs"* so it doesn't imply sBPF coverage.

Deferring the `--list-rules` passthrough and the `ratchet-anchor` `default-features` peel as you suggested — happy to open separate PRs for both after this lands.

All four checks green on the follow-up commit: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (167 tests total, +3 from the template snapshots), `scripts/check-readme-drift.sh`.

PR #7, opened 2026-04-23
feat(qed): address deferred review items from #6 — list-rules + dep trim

saicharanpogul:
## Summary

Follow-up to [#6](https://github.com/QEDGen/solana-skills/pull/6). Addresses the two items you flagged as deferred in the review:

1. **`ratchet list-rules` → `qedgen readiness --list-rules` / `qedgen check-upgrade --list-rules`.** The README referenced the upstream CLI, but qedgen embeds ratchet as a library — users installing via `install.sh` / `npx skills add` don't have the `ratchet` binary on PATH. Now the catalog is discoverable through qedgen itself.
2. **`curve25519-dalek` + `fiat-crypto` dropped from qedgen's dep tree.** Published `ratchet 0.3.1` this morning that gates those behind the `rpc` feature; bumping the pin here picks them up.

### `qedgen readiness --list-rules`
Prints the 6-rule P-catalog (human table to stderr). Pair with `--json` for a machine-readable dump to stdout.

```
$ qedgen readiness --list-rules
qedgen readiness (preflight, P-rules) — 6 rule(s):
  P001  account-missing-version-field             Accounts without a leading `version: u8` have no in-band signal for layout-version branching…
  P002  account-missing-reserved-padding          Accounts without a trailing `_reserved: [u8; N]` padding require a realloc or migration on every future field append.
  P003  account-missing-discriminator-pin         Account uses Anchor's default discriminator; pinning it explicitly survives struct renames without firing R006.
  P004  event-missing-discriminator-pin           Event uses Anchor's default discriminator; pinning it survives renames without desyncing indexers.
  P005  account-name-collision                    Account struct is named after a well-known Solana type, making tooling and docs ambiguous.
  P006  instruction-missing-signer                Instruction writes to accounts with no signer slot — anyone can mutate state unless CPI authority is checked elsewhere.
```

### `qedgen check-upgrade --list-rules`
Same shape for the 16-rule R-catalog.

### CLI shape
`--idl` / `--old` / `--new` become optional via clap's `required_unless_present = "list_rules"`, so the happy-path signatures are unchanged:
- `qedgen readiness --idl x.json` — still works exactly as before
- `qedgen readiness --list-rules` — skips the engine, dumps the catalog, exit 0

### Ratchet 0.3.1 dep bump
`ratchet-anchor 0.3.1` gates `curve25519-dalek` (and its `fiat-crypto` sub-tree) behind the `rpc` feature. qedgen sets `default-features = false`, so those crates were dead weight in the dep graph since #6. Verified dropped:

```
$ cargo tree -p qedgen-solana-skills | grep -c curve25519
0
$ cargo tree -p qedgen-solana-skills | grep -c fiat-crypto
0
```

Release notes + publish log: https://github.com/saicharanpogul/ratchet/releases/tag/v0.3.1

No behavioural change — the default ratchet CLI and any consumer with `default-features = true` keeps the old dep path. Patch-level bump, non-breaking.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace` — 222 tests total, +2 new (`list_rules_preflight_covers_full_p_catalog`, `list_rules_diff_covers_full_r_catalog`) lock the rule-catalog shape so a future upstream bump that drops or renames a rule surfaces loudly
- [x] `bash scripts/check-readme-drift.sh` — 15/15 subcommands documented
- [x] End-to-end: `qedgen readiness --list-rules` (human), `qedgen check-upgrade --list-rules --json` (machine) both produce correct output
- [x] `cargo tree` dep-drop verification above

## Files touched

| File | Change |
|---|---|
| `crates/qedgen/Cargo.toml` | `ratchet-core` / `ratchet-anchor` pin `"0.3"` → `"0.3.1"` |
| `crates/qedgen/src/main.rs` | `--list-rules` flag on `Readiness` + `CheckUpgrade`, `--idl` / `--old` / `--new` made conditional, early-exit match-arm path |
| `crates/qedgen/src/ratchet.rs` | `print_rules_preflight()` / `print_rules_diff()` + `RuleEntry` serde struct + 2 catalog-shape snapshot tests |
| `README.md` | Swap `ratchet list-rules` reference for the new `qedgen readiness --list-rules` / `check-upgrade --list-rules` flags |
| `Cargo.lock` | `solana-ratchet-{core,anchor}` pulled in at 0.3.1; `curve25519-dalek` + `fiat-crypto` drop out of qedgen's sub-graph |

Total: ~136 net LOC added, 5 files. Same themed-bundle cadence as #6.

## Also addressed from #6's non-blocking notes

These were already fixed in #6's follow-up commit (`926f34e`):
- Stderr/stdout comment on `print_human` / `print_json`
- Trailing blank line in regenerated `verify.yml`
- README "Deploy safety" row qualified as Anchor-only
- Exit code 3 wired for caller-side errors
- Template-expansion snapshot tests

Nothing outstanding from that review as far as I can tell. Happy to resize or split this PR if either piece warrants a separate review thread.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,32 +247,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "curve25519-dalek"
-version = "4.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97fb8b7c4503de7d6ae7b42ab72a5a59857b4c937ec27a3d4539dba95b5ab2be"
-dependencies = [
- "cfg-if",
- "cpufeatures",
- "curve25519-dalek-derive",
- "fiat-crypto",
- "rustc_version",
- "subtle",
- "zeroize",
-]
-
-[[package]]
-name = "curve25519-dalek-derive"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -314,12 +288,6 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
-
-[[package]]
-name = "fiat-crypto"
-version = "0.2.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28dea519a9695b9977216879a3ebfddf92f1c08c05d984f8996aecd6ecdc811d"
 
 [[package]]
 name = "filetime"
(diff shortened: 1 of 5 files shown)


abishekk92:
Looks good to me, thanks for your contribution

PR #9, opened 2026-04-24
fix(typecheck): reject Pubkey vs named numeric const (F7+F8 follow-up)

tanmay4l:


abishekk92 (approved):
LGTM — gates clean, F7/F8 tests cover the three cases that matter (assign, compare, negative guard that Pubkey params still work). Safe to merge.

One minor follow-up (not blocking): `collect_numeric_consts` walks only top-level `spec.items`, not `TopItem::Pragma(p).items`. So a `pragma sbpf { const ZERO = 0 }` that's later used against a Pubkey field would slip through. The handler DSL this check targets doesn't currently live under pragmas, so it's theoretical — but a one-line recursive descent would close it. Fine as a follow-up PR if you want this one to stay focused.

abishekk92:
Thanks for the follow-up on F7/F8 — the extension to named numeric consts is the obvious next step and the test coverage (assign, compare, param-still-valid) is clean.

One ask before merge: could you add matching `.qedspec` regression fixtures under `examples/regressions/issue-8/` to mirror the existing `repro-07-pubkey-literal-assign.qedspec` / `repro-08-pubkey-literal-compare.qedspec` pattern? Something like:

- `repro-07-const-pubkey-assign.qedspec` — `const ZERO = 0` + `effect { key := ZERO }`
- `repro-08-const-pubkey-compare.qedspec` — `const ZERO = 0` + `requires state.key != ZERO`

Each with the same inline-header-comment style (finding number, expected-vs-actual, fix site). Those fixtures are the ratchet layer above the Rust unit tests — when we wire CI to run `qedgen check` over `examples/regressions/**/*.qedspec`, they'll gate any future regression.

No other blockers — happy to merge once the fixtures land.

tanmay4l:
> Thanks for the follow-up on F7/F8 — the extension to named numeric consts is the obvious next step and the test coverage (assign, compare, param-still-valid) is clean.
> 
> One ask before merge: could you add matching `.qedspec` regression fixtures under `examples/regressions/issue-8/` to mirror the existing `repro-07-pubkey-literal-assign.qedspec` / `repro-08-pubkey-literal-compare.qedspec` pattern? Something like:
> 
> * `repro-07-const-pubkey-assign.qedspec` — `const ZERO = 0` + `effect { key := ZERO }`
> * `repro-08-const-pubkey-compare.qedspec` — `const ZERO = 0` + `requires state.key != ZERO`
> 
> Each with the same inline-header-comment style (finding number, expected-vs-actual, fix site). Those fixtures are the ratchet layer above the Rust unit tests — when we wire CI to run `qedgen check` over `examples/regressions/**/*.qedspec`, they'll gate any future regression.
> 
> No other blockers — happy to merge once the fixtures land.

added both fixtures under examples/regressions/issue-8/  same inlineheader style as the existing ones. The pragma recursion follow-up is noted, will handle in a separate PR.

abishekk92:
Verified end-to-end:

- 244 tests pass (3 new), `cargo fmt`/`clippy` clean
- Both fixtures (`repro-07-const-pubkey-assign.qedspec`, `repro-08-const-pubkey-compare.qedspec`) exercise the new typecheck correctly: `qedgen check` rejects `key := ZERO` and `state.key != ZERO` with the expected Pubkey-vs-numeric-const error messages

One procedural ask before merge: please rebase onto `main`. The branch predates 70f4ca9 (the issue-template commit that landed yesterday), so the diff currently shows `.github/ISSUE_TEMPLATE/qedspec-bug.yml` as a deletion — a clean merge would revert the template. After rebase this is good to go.

PR #11, opened 2026-04-24
feat: implement forall/exists quantifier support

tanmay4l:
Implements quantifier support for `forall`/`exists` in `.qedspec` properties.

- U8/I8 quantifiers now lower to real Rust iterators (for example, `(u8::MIN..=u8::MAX).all(|v| …)`), so they no longer silently pass.
- U16+ quantifiers now emit a P1 `unchecked_quantifier` lint via `qedgen check` instead of silently returning `true`.
- Fixed Lean def emission that previously stripped `∀ v :` binders and left `v` unbound.
- Preservation proof templates now emit `sorry` for quantified properties instead of `omega`, since `omega` cannot prove universal goals and was generating non-compiling Lean.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -763,10 +763,17 @@ Accepts comma- or newline-separated fields; desugars to
 alongside the multi-line form.
 
 **`implies`** in property bodies lowers to `(!a) || (b)` in generated
-Rust (Kani/proptest). **`forall` / `exists`** in a property body emit a
-`/* QEDGEN_UNSUPPORTED_QUANTIFIER */` marker — the property fn returns
-`true` and you should express the quantification at the harness level
-via `kani::any()` / proptest generators.
+Rust (Kani/proptest). **`forall` / `exists`** over **U8 or I8** lower to
+`(T::MIN..=T::MAX).all(|v| body)` / `.any(|v| body)` — exhaustive check,
+≤256 iterations, works in both proptest and Kani. Larger types (U16+)
+still emit a `/* QEDGEN_UNSUPPORTED_QUANTIFIER */` sentinel and the
+property fn returns `true`; `qedgen check` fires a P1
+`unchecked_quantifier` warning so the gap is visible rather than silent.
+
+**Quantified-property preservation theorems** (property body contains
+`∀`/`∃`) default to `sorry` in generated Lean — `omega` cannot prove
+universal goals. Fill these via the standard escalation ladder:
+Leanstral first (`qedgen fill-sorry`), then Aristotle for harder goals.
 
 **Multi-variant ADTs with shared field names** deduplicate when
 flattened into the state model (first-variant wins on name collisions).
--- a/crates/qedgen/src/check.rs
+++ b/crates/qedgen/src/check.rs
@@ -1793,6 +1793,47 @@ pub fn check_completeness(spec: &ParsedSpec) -> Vec<CompletenessWarning> {
         }
     }
 
+    // Rule: quantifier over a type that can't be exhausted at test time.
+    // U8/I8 are fine (256 iterations); anything larger emits a stub `true`
+    // body, so the property silently passes without being checked.
+    for prop in &spec.properties {
+        if let Some(ref rust_expr) = prop.rust_expression {
+            if rust_expr_is_unsupported(rust_expr) {
+                // Extract the quantifier kind and binder type from the sentinel
+                // comment so the message is specific.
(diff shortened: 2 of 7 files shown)


abishekk92:
Solid PR — the tiered approach (U8/I8 lower to real iterators, U16+ emit the sentinel + P1 lint) is the right shape, and the `render_properties_inner` strip fix catches a genuine bug (stripping `∀ v :` would leave `v` unbound).

Two asks before merge:

**1. Regression fixtures.** Could you add `.qedspec` files under `examples/regressions/` covering:

- `repro-forall-u8.qedspec` — positive: U8 quantifier, expect no lint, expect a valid `.all(|v| …)` in the generated proptest harness
- `repro-forall-u64.qedspec` — negative: U64 quantifier, expect the new `unchecked_quantifier` P1 warning
- `repro-forall-value-binder.qedspec` — regression guard for the Lean `∀ v :` strip fix (spec with `forall v : U8, v >= 0` that previously produced `:= v ≥ 0` with `v` unbound)

Same inline-header-comment convention as `examples/regressions/issue-8/`.

**2. `sorry` in preservation proofs.** `preservation_proof_script` now emits `sorry` for quantified properties. Net-positive vs non-compiling `omega`, but it does introduce `sorry` into the generated Lean — which the [pre-release checklist](https://github.com/QEDGen/solana-skills/blob/main/CLAUDE.md#L438) currently gates on (`grep -r '\bsorry\b' examples/**/*.lean` must be empty). No existing example uses quantifiers so nothing breaks today, but it's worth a short note in `SKILL.md` (or wherever the escalation ladder lives) that quantified-property preservation defaults to the Leanstral/Aristotle lane. Not a blocker for this PR — just making sure the expectation is set.

Happy to iterate.

tanmay4l:
> Solid PR — the tiered approach (U8/I8 lower to real iterators, U16+ emit the sentinel + P1 lint) is the right shape, and the `render_properties_inner` strip fix catches a genuine bug (stripping `∀ v :` would leave `v` unbound).
> 
> Two asks before merge:
> 
> **1. Regression fixtures.** Could you add `.qedspec` files under `examples/regressions/` covering:
> 
> * `repro-forall-u8.qedspec` — positive: U8 quantifier, expect no lint, expect a valid `.all(|v| …)` in the generated proptest harness
> * `repro-forall-u64.qedspec` — negative: U64 quantifier, expect the new `unchecked_quantifier` P1 warning
> * `repro-forall-value-binder.qedspec` — regression guard for the Lean `∀ v :` strip fix (spec with `forall v : U8, v >= 0` that previously produced `:= v ≥ 0` with `v` unbound)
> 
> Same inline-header-comment convention as `examples/regressions/issue-8/`.
> 
> **2. `sorry` in preservation proofs.** `preservation_proof_script` now emits `sorry` for quantified properties. Net-positive vs non-compiling `omega`, but it does introduce `sorry` into the generated Lean — which the [pre-release checklist](https://github.com/QEDGen/solana-skills/blob/main/CLAUDE.md#L438) currently gates on (`grep -r '\bsorry\b' examples/**/*.lean` must be empty). No existing example uses quantifiers so nothing breaks today, but it's worth a short note in `SKILL.md` (or wherever the escalation ladder lives) that quantified-property preservation defaults to the Leanstral/Aristotle lane. Not a blocker for this PR — just making sure the expectation is set.
> 
> Happy to iterate.

added the three fixtures (repro-forall-u8, repro-forall-u64, repro-forall-value-binder) and the SKILL.md note pointing quantified preservation to the Leanstral/Aristotle lane.

abishekk92:
Verified all three fixtures end-to-end on the branch:

- `repro-forall-u8`: proptest emits `(u8::MIN..=u8::MAX).all(|v| v >= 0)`, Lean def preserves `∀ v : Nat, v ≥ 0`, no quantifier warning fires
- `repro-forall-u64`: proptest emits `true` stub, `qedgen check` fires the P1 `unchecked_quantifier` lint naming the property — the silent-skip is now visible
- `repro-forall-value-binder`: Lean def preserves the binder; the strip-bug regression guard works

246 tests pass (5 new), `cargo fmt`/`clippy` clean. SKILL.md note about quantified-preservation defaulting to `sorry` + Leanstral/Aristotle escalation is in place. Existing examples don't use quantifiers so the zero-sorry pre-release gate is unaffected.

Same rebase ask as #9 — branch predates 70f4ca9 (issue template), so please rebase onto `main` before merge. Otherwise good to go.

Issue #8, opened 2026-04-24
v2.7.0 field eval: 8 Lean codegen / check-time findings from a real-world Anchor spec


lmvdz:
## Context

Field-evaluated v2.7.0 against a production Anchor program with
~13 handlers, ~29 error variants, and a folded-state model that unifies
several Anchor account types. Shipped at **0 `!` warnings, 6 `i [P2]
[no_lifecycle]` info** at `qedgen check` (all 6 are deliberate pure-CPI
handlers). `qedgen codegen --lean` + `lake build` on Lean 4.15.0 /
Lake 5.0.0 surfaces **8 distinct breakages**, all reproducible with
`< 10 line` standalone specs — 7 are Lean-codegen bugs, 1 is a
check-time type-validation gap.

None are covered by the bundled examples because the examples don't
exercise the triggering feature combinations (see per-finding "Why
untested" notes — most boil down to "bundled specs don't combine `or`
in requires, uninterpreted helpers in guards, parameterized liveness,
or Bool state fields").

## TL;DR

| # | Symptom | Root cause (source line) |
|---|---|---|
| 1 | `U16`/`U32`/`I8..I64` emit as literal type name, undefined in Lean | `lean_gen::map_type` at [lean_gen.rs:2748-2754](https://github.com/QEDGen/solana-skills/blob/v2.7.0/crates/qedgen/src/lean_gen.rs#L2748) missing mappings that Rust's `codegen::map_type` got in v2.6.2 S1 |
| 2 | Composite guards mis-paren `∨`/`→` — Lean parses `A ∧ B ∨ C` as `(A ∧ B) ∨ C` | [lean_gen.rs:367-369](https://github.com/QEDGen/solana-skills/blob/v2.7.0/crates/qedgen/src/lean_gen.rs#L367) pushes `req.lean_expr` raw then ∧-joins downstream, no paren-wrap |
| 3 | Two `requires X else SameErr` → duplicate theorem `h_aborts_if_SameErr` | [lean_gen.rs:1897](https://github.com/QEDGen/solana-skills/blob/v2.7.0/crates/qedgen/src/lean_gen.rs#L1897) formats `{op}_aborts_if_{error}` with no dedup |
[...]

PR #13, opened 2026-04-25
fix(lean_gen): emit uninterpreted helpers as `opaque … : Bool` (issue #12)

### LINKED ISSUE #12: v2.7.1 F5 followup: uninterpreted helper axioms are `→ Prop`, can't be used in `requires` (computable if-guard needs Decidable)
Followup to issue #8, F5. v2.7.1 added `axiom` emission for uninterpreted helpers — nice fix, but they land as `→ Prop` which means `lake build` rejects any spec that uses them in `requires` position.

## Minimal repro (13 lines)

```qedspec
spec ReproF5
program_id "11111111111111111111111111111111"

type State
  | Active of { c : U8 }

type Error | E

handler h : State.Active -> State.Active {
  permissionless
  requires foo(state.c) else E
  effect { c := 1 }
}
```

```bash
QEDGEN="$HOME/.agents/skills/qedgen/tools/qedgen"
mkdir repro && cd repro && git init -q
cp spec.qedspec .
"$QEDGEN" init --name ReproF5 --spec spec.qedspec --output-dir .
"$QEDGEN" codegen --lean --lean-output Spec.lean
lake build
```

Result:
```
axiom foo : Nat → Prop                     -- line 15 of generated Spec.lean

error: Spec.lean:27:2: failed to synthesize instance of type class
  Decidable (s.status = Status.Active ∧ foo s.c)
```

## Why

`requires` clauses lower to the if-guard of the transition function:
```lean
def hTransition (s : State) (signer : Pubkey) : Option State :=
  if s.status = Status.Active ∧ foo s.c then  -- ← needs Decidable
    some { s with c := 1 }
  else none
```
Lean needs `Decidable (P)` to compile `if P then ... else ...` in a
`def` body. `axiom foo : Nat → Prop` gives an opaque `Prop` with no
`Decidable` instance.

## What I tried as a user-side workaround

Adding `open Classical` at the top of generated `Spec.lean` moves
past the Decidable error, but then:
```
error: failed to compile definition, consider marking it as 'noncomputable'
  because it depends on 'propDecidable', which is 'noncomputable'
```

Transition functions have to be computable (they're used under `applyOp`
/ `applyOps` pattern matching), so classical decidability isn't a viable
escape hatch. The fix has to come from the axiom-emission side.

## Fix options

**(a) Emit axioms as `→ Bool` instead of `→ Prop`.** Bool is `Decidable`
by construction; `if foo s.c then ...` works naturally. Tradeoff:
`Bool`-valued uninterpreted helpers are slightly less idiomatic for
proof-style reasoning than `Prop`-valued ones, but given these are
explicitly *uninterpreted* (placeholder for a real check or a hand-
written proof step), the convenience wins.

**(b) Emit a paired `Decidable` instance alongside each axiom.**
```lean
axiom foo : Nat → Prop
instance (n : Nat) : Decidable (foo n) := Classical.propDecidable _
```
Works at the typeclass level but introduces `propDecidable` into the
transition-function call path, which Lean then flags as noncomputable
for the same reason my open-Classical probe failed. Unless the whole
transition is marked `noncomputable`, which breaks `applyOp` pattern
matching downstream.

**(c) Keep axiom as `→ Prop` but also generate a `Bool`-valued
companion that transitions use.**
```lean
axiom foo : Nat → Prop         -- for proofs / ensures / properties
opaque fooB : Nat → Bool       -- for requires lowering
axiom fooB_reflects : ∀ n, fooB n = true ↔ foo n
```
Proofs can reason in `Prop`; transitions compute in `Bool`. Heavier
scaffolding but preserves the proof-ergonomic `Prop` axiom.

My lean is **(a)** — shortest diff, matches the "uninterpreted helper
is a black box I'll define later" mental model, and the `Bool →`
form lets codegen emit `requires` *and* `ensures`/`property` positions
identically. **(c)** is theoretically richer but only pays off if
downstream proofs routinely want `Prop`-form access, which isn't my
use case.

## Impact on real specs

On my 13-handler spec (see issue #8 fixtures), 6 distinct call sites
want uninterpreted-helper guards surfacing:
- `input_handles_equal_stored(sp)` — stale-CT rollback prevention on
  submit / update / match / request-withdraw (4 sites)
- `account_id_ct_binds_signer(admin)` — concern-30 signer binding on
  read-balance and request-withdraw
- `graph_bytes_matches_pin(kind, bytes)` — Phase-0 hash-pin on
  register-graph

All are blocked on this. The guards live as documentation-only
invariants in the spec and don't get Kani / proptest coverage. F5's
whole promise was to surface these as machine-checkable; today they
land at `qedgen check` but break at `lake build`.

## Environment

- qedgen v2.7.1 (release binary, sha256-verified)
- Lean 4.15.0 / Lake 5.0.0-1165156

Happy to PR (a) if you're open to `Bool`-valued axioms — looks like a
~10-line change in `lean_gen.rs` where F5 landed (commit 3db0a1c).

lmvdz:
## Summary

Closes #12. v2.7.1 F5 emitted uninterpreted helpers as `axiom foo : T → Prop` — that fixed the unknown-identifier failure from issue #8 finding #5, but left the helper unusable in `requires` position. That clause lowers to a transition function's `if`-guard which Lean requires to be `Decidable`, and `axiom … → Prop` is opaque + noncomputable.

This PR emits `opaque foo : T → Bool` instead. Two changes ripple from it:

- **`Bool` instead of `Prop`** — `Bool` is auto-`Decidable`, so the `if`-guard typechecks naturally. Call-site rendering (`(foo (y))`) is unchanged; the elaborator's `Bool ⇒ Prop` coercion (`b = true`) handles `ensures` / `property` positions.
- **`opaque` instead of `axiom`** — `opaque` declares an irreducible body via the type's `Inhabited` instance (`Bool` defaults to `false`), keeping the transition function computable. `axiom` would force `noncomputable` on every transition that references the helper, which then breaks `applyOp` pattern-matching downstream.

Spec authors who want to discharge a helper rather than trust it still upgrade to a real `def` in their support module — same path as before, just one keyword different.

## Diff (~78 insertions, 34 deletions, 3 files)

| File | Change |
|---|---|
| `crates/qedgen/src/chumsky_adapter.rs` | `walk_apps` push return type `"Prop"` → `"Bool"`. One-line fix; comment expanded explaining why. |
| `crates/qedgen/src/lean_gen.rs` | `emit_uninterpreted_helpers` emits `opaque` instead of `axiom`; doc comment updated to reflect the new shape and the rationale (irreducible body via `Inhabited`, transitions stay computable). Test `finding_5_uninterpreted_helpers_are_axiomatized` → `finding_5_uninterpreted_helpers_are_opaque_bool`, with positive assertion on the new shape and negative assertions on the old `axiom` / `→ Prop` shapes to prevent regression. |
| `examples/regressions/issue-8/repro-05-uninterpreted-helper.qedspec` | Inline header doc updated to record the v2.7.0 → v2.7.1 → v2.7.2 progression. |

## Test plan

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test --workspace` — 241 passed, 0 failed (including the renamed `finding_5_uninterpreted_helpers_are_opaque_bool`)
- [x] All 10 bundled `examples/{rust,sbpf}/*/formal_verification/` regen + `lake build` clean (none use App-position helpers, so the change is structurally additive for them — verified no Lean elaboration paths broke)
- [x] `examples/regressions/issue-8/repro-05-uninterpreted-helper.qedspec` end-to-end: `qedgen init` → `qedgen codegen --lean` → `lake build` succeeds (this was the original repro from issue #12; failed under v2.7.1)
- [x] Live spec from my dagon work — 13 handlers, 6 distinct App-position uninterpreted helpers in `requires` positions, 1773-line generated `Spec.lean` — also `lake build` clean post-patch (not included in this PR but a reasonable smoke test on the size of spec the change is supposed to unlock)

## Why now

Issue #12 was filed yesterday with the minimal repro and three fix-direction options. You replied that option (a) — `→ Bool` axioms — looked straightforward; I'm submitting (a) refined to `opaque … : Bool` because `axiom … : Bool` still hits the noncomputability problem (axioms-as-keyword force noncomputable, regardless of return type). Same surface API — the existing `infer_lean_type` helper continues to drive arg types — just the declaration kind moves.

### DIFF
--- a/crates/qedgen/src/chumsky_adapter.rs
+++ b/crates/qedgen/src/chumsky_adapter.rs
@@ -1347,7 +1347,17 @@ fn walk_apps(
                     .iter()
                     .map(|n| infer_lean_type(&n.node, field_types, param_types))
                     .collect();
-                out.push((func.clone(), arg_types, "Prop".to_string()));
+                // Bool, not Prop. The original v2.7.1 F5 emission used
+                // `→ Prop`, which lands at codegen but breaks `lake build`
+                // — `requires` / `ensures` clauses lower to a transition
+                // function's `if`-guard, which Lean requires to be
+                // `Decidable`. `axiom foo : T → Prop` is opaque and
+                // noncomputable, so the transition fails to compile.
+                // `Bool` is auto-`Decidable` and lifts cleanly into
+                // `Prop` positions via the standard `b = true` coercion
+                // that the call-site renderer already produces. See
+                // issue #12.
+                out.push((func.clone(), arg_types, "Bool".to_string()));
             }
             for n in args {
                 walk_apps(&n.node, field_types, param_types, out, seen);
--- a/crates/qedgen/src/lean_gen.rs
+++ b/crates/qedgen/src/lean_gen.rs
@@ -399,18 +399,31 @@ fn build_guard_cond_parts(
 /// from the spec. These are functions referenced by name in guard /
 /// ensures / effect / property bodies but never defined structurally —
 /// user-facing "named but-not-fully-modeled" security check helpers.
-/// Issue #8 finding #5. Without this emission, generated Lean references
-/// an identifier Lake never sees declared.
+/// Issue #8 finding #5 (initial axiom emission), issue #12 (use
+/// `opaque` rather than `axiom` so transitions stay computable).
+///
+/// `opaque` rather than `axiom`: an `axiom`'s declared identifier is
+/// permanently noncomputable, which propagates to any `def` that
+/// references it — including the per-handler transition functions
+/// generated below. `opaque T` declares a top-level definition whose
+/// body is hidden but is computable via the type's `Inhabited`
+/// instance (`Bool` is auto-`Inhabited`, defaulting to `false`), so
+/// the `if`-guard inside a transition function compiles. Users who
+/// want to strengthen a helper into a real check replace the `opaque`
+/// declaration with a `def` in their `Proofs.lean` (or a sibling
+/// support module imported before `Spec.lean`).
 fn emit_uninterpreted_helpers(out: &mut String, helpers: &[(String, Vec<String>, String)]) {
     if helpers.is_empty() {
         return;
     }
     out.push_str(
-        "-- Uninterpreted helpers: declared axiomatically so generated\n\
-         -- proofs typecheck even though the DSL doesn't model their\n\
-         -- semantics. Treat each as an abstract property; strengthen\n\
-         -- into a concrete definition in your support module if you\n\
-         -- want to discharge it (rather than trust it).\n",
+        "-- Uninterpreted helpers: declared opaquely so generated\n\
+         -- transitions typecheck even though the DSL doesn't model\n\
+         -- their semantics. Treat each as an abstract Bool predicate;\n\
+         -- strengthen into a concrete definition in your support\n\
+         -- module if you want to discharge it (rather than trust it).\n\
+         -- `opaque` keeps the transition functions computable\n\
+         -- (axioms would force them noncomputable).\n",
     );
     for (name, arg_types, return_type) in helpers {
         let sig = if arg_types.is_empty() {
@@ -420,7 +433,7 @@ fn emit_uninterpreted_helpers(out: &mut String, helpers: &[(String, Vec<String>,
             parts.push(return_type.clone());
             parts.join(" \u{2192} ")
         };
-        out.push_str(&format!("axiom {} : {}\n", safe_name(name), sig));
+        out.push_str(&format!("opaque {} : {}\n", safe_name(name), sig));
     }
     out.push('\n');
 }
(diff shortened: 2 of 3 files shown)


abishekk92:
Verified the test plan end-to-end on a fresh checkout:

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (241 passed) — all green
- Renamed test `finding_5_uninterpreted_helpers_are_opaque_bool` runs and passes; the negative assertions guarding against the `axiom` / `→ Prop` shape are load-bearing
- End-to-end smoke test on the issue #12 minimal repro (`init` → `codegen --lean` → `lake build`) — succeeds, 8 jobs. Generated Lean has `opaque foo : Nat → Bool` followed by a computable `def hTransition` exactly as designed
- Branch is rebased on current main

`opaque … : Bool` is the right call — `axiom … : Bool` would still hit the noncomputability problem you flagged. Ready to merge from my side.

PR #20, opened 2026-04-27
feat(qed): extend ratchet integration to cover Quasar IDLs

saicharanpogul:
## Summary

Continuation of #6 / #7. Extends the embedded ratchet integration so `qedgen readiness` and `qedgen check-upgrade` work against [Quasar](https://github.com/blueshift-gg/quasar)-emitted IDLs in addition to Anchor — same subcommands, same exit-code contract, same rule catalog. Picks up `solana-ratchet-quasar 0.3.2` as a third dep alongside the existing `core` + `anchor` pins.

The framework-agnostic R/P rule engine downstream is unchanged: both loaders lower into ratchet's `ProgramSurface` IR before any rule fires.

### `--quasar` flag + `Quasar.toml` autodetect

```bash
# Explicit
qedgen readiness --idl target/idl/escrow.json --quasar
qedgen check-upgrade --old ratchet.lock --new target/idl/escrow.json --quasar

# Implicit — pick up Quasar when `Quasar.toml` lives in the cwd and no
# `Anchor.toml` shadows it. Mixed projects fall back to Anchor.
cd my-quasar-program
qedgen readiness --idl target/idl/escrow.json
# → eprintln: \`qedgen: Quasar project detected (Quasar.toml in cwd) — using ratchet's Quasar IDL parser\`
```

Autodetect's stderr banner is suppressed under `--json` to keep stdout machine-parseable.

### Worked example: `examples/quasar-readiness/`

Two committed Quasar IDL fixtures (`escrow.json` + `escrow.v2.json`) shaped exactly like `quasar build` writes to `target/idl/<program>.json`. The v1 is hackathon-typical (no `version`, no `_reserved`) so readiness flags it; the v2 is a deliberately broken upgrade firing R001 / R006 / R007 / R008 at once. README walks through both runs and the autodetect path.

Quasar's compiler isn't on qedgen's CI, so we commit JSON shaped to match its emission rather than building it. Same convention `solana-ratchet-quasar` follows in its own test fixtures.

### Three new tests locking the dispatch

In `crates/qedgen/src/ratchet.rs`:

- `quasar_readiness_flags_bare_v1_surface` — runs P-rules on a bare Quasar v1 surface; asserts P001 + P002 fire and P003 + P004 do **not** (sha256-default rules are a category error on Quasar — devs always pin discriminators in source, so the rule has nothing to flag).
- `quasar_check_upgrade_catches_discriminator_change` — flips the `Escrow` discriminator from `[42]` → `[99]` and asserts R006 fires.
- `quasar_anchor_idl_under_quasar_mode_is_a_parse_error` — feeds an Anchor-shaped IDL through `Framework::Quasar`. Confirms the dispatch is wired to the Quasar parser rather than silently falling back to Anchor.

The pre-existing `Framework::Anchor` test path stays exercised — every Anchor case has its `framework: Framework::Anchor` field threaded through.

## Why now

ratchet 0.3.2 (published 2026-04-27) ships first-class Quasar support: a parallel parser for Quasar's IDL JSON shape (1-byte variable-length account discriminators, untagged \`IdlType\` union, struct-only typedefs joined to accounts by name) that lowers into the same \`ProgramSurface\` the existing R/P rules already operate on. With the ratchet side stable, qedgen can pick it up at zero rule-count cost.

The bigger picture: qedgen already advertises Quasar as a codegen target in v2.9+ (\`qedgen init --target quasar\`). Until this PR, the deploy-safety lint silently broke that promise — Quasar users got a parse error from the embedded ratchet anchor loader. Now the README's \"Anchor and Quasar\" framing holds end-to-end.

A potential future follow-up — outside this PR's scope — would be a Quasar-side compile-time linter that runs the same rule set natively against the AST instead of going through the IDL JSON. That avoids the round-trip and could surface findings before \`quasar build\` writes anything to disk. Happy to scope that in a separate discussion if there's interest.

## Files touched

| File | Change |
|---|---|
| \`crates/qedgen/Cargo.toml\` | Add \`ratchet-quasar = { package = \"solana-ratchet-quasar\", version = \"0.3.2\" }\`; bump \`ratchet-core\` / \`ratchet-anchor\` pins to 0.3.2 |
| \`crates/qedgen/src/ratchet.rs\` | New \`Framework\` enum + \`detect_from_cwd\`; \`load_surface(path, framework)\` dispatch; +3 Quasar dispatch tests; existing Anchor tests pass through \`framework: Framework::Anchor\` |
| \`crates/qedgen/src/main.rs\` | \`--quasar\` flag on \`Readiness\` + \`CheckUpgrade\`; \`resolve_framework(explicit, as_json)\` helper for autodetect-with-banner |
| \`README.md\` | \"Deploy safety\" row + lint section reframed as Anchor + Quasar; example pointer added |
| \`examples/quasar-readiness/\` | New: \`escrow.json\` + \`escrow.v2.json\` + README walk-through |
| \`Cargo.lock\` | \`solana-ratchet-quasar 0.3.2\` resolves into the workspace; no transitive surprises |

Total: ~480 net LOC added, 8 files. Same themed-bundle cadence as #6 / #7.

## Test plan

- [x] \`cargo fmt --all -- --check\`
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\`
- [x] \`cargo test --workspace\` — 442 tests total, +3 new (\`quasar_readiness_flags_bare_v1_surface\`, \`quasar_check_upgrade_catches_discriminator_change\`, \`quasar_anchor_idl_under_quasar_mode_is_a_parse_error\`)
- [x] \`bash scripts/check-readme-drift.sh\` — 17/17 subcommands documented
- [x] End-to-end: \`qedgen readiness --idl examples/quasar-readiness/escrow.json --quasar\` flags P001 + P002 and exits 2; \`qedgen check-upgrade --old escrow.json --new escrow.v2.json --quasar\` flags R001 + R006 + R007 + R008 and exits 1
- [x] Catalog count guards still pin 22 rules (6 P + 16 R) — Quasar is dispatch-only, doesn't add rules
- [x] No off-graph crate adds: \`cargo tree -p qedgen-solana-skills\` shows only \`solana-ratchet-{core,anchor,quasar}\` from the ratchet family


abishekk92 (commented):
Solid PR — design is clean and the test coverage punches above its weight. Two doc-only nits worth fixing before merge; neither blocks behavior, neither was catchable by CI.

## What's strong

- **Dispatch is well-factored.** `Framework` enum at the seam, single `load_surface(path, framework)` matching on it. Both branches lower into the same `ProgramSurface`, so the rule engine downstream is identical — no rule duplication, no silent divergence.
- **Test coverage earns each test.** `quasar_readiness_flags_bare_v1_surface` asserts P001/P002 fire **and** P003/P004 *don't* — the negative assertion is the one that catches "we silenced everything by accident." `quasar_anchor_idl_under_quasar_mode_is_a_parse_error` is exactly the test I'd have asked for if it weren't there: proves dispatch reaches the Quasar parser instead of silently falling back to Anchor.
- **No regression risk on the Anchor path** — every pre-existing test threads `framework: Framework::Anchor` through.
- **`resolve_framework(explicit, as_json)` wiring is right**: explicit wins, banner fires only on Quasar autodetect, suppressed under `--json` so stdout stays machine-parseable.

## Nit 1 — `detect_from_cwd` docstring lies about parent walking

`crates/qedgen/src/ratchet.rs` lines 48–57:

\`\`\`rust
/// Walk the current working directory's parents looking for a
/// project marker — \`Anchor.toml\` or \`Quasar.toml\`.
pub fn detect_from_cwd() -> Self {
    if std::path::Path::new(\"Quasar.toml\").exists()
        && !std::path::Path::new(\"Anchor.toml\").exists()
\`\`\`

The implementation only checks the cwd, not parents. Cargo-style tooling lets devs run from anywhere in the workspace — that's what the docstring promises but isn't what's wired. Either:

- Drop \"parents\" from the docstring if cwd-only is intentional, or
- Walk up via \`std::env::current_dir()?.ancestors()\` if parent-walking was the intent (likely correct long-term — hackathon devs run \`qedgen\` from a nested dir).

Doc-only fix is fine for the merge; full implementation can be a follow-up.

## Nit 2 — Test comment misdescribes the Quasar parse failure

\`crates/qedgen/src/ratchet.rs\` lines 496–499:

\`\`\`rust
// Anchor IDLs use a \`type: { kind: \"struct\", fields: ... }\` shape
// for typedefs; Quasar's loader rejects that.
\`\`\`

The \`QUASAR_BARE_V1_IDL\` fixture above (lines 432–440) uses *exactly that shape*:

\`\`\`json
\"type\": { \"kind\": \"struct\", \"fields\": [{ \"name\": \"balance\", \"type\": \"u64\" }] }
\`\`\`

So Quasar clearly accepts struct typedefs. The actual reason the Anchor IDL fails under Quasar mode is almost certainly the **discriminator shape** (Anchor's \`[u8; 8]\` vs Quasar's 1-byte variable-length). Test assertion is correct; comment is just wrong. One-line fix.

## Lower-priority

- \`Cargo.toml\` block comment mentions \"two aliases\" — three now. Optional touch-up.
- Mixed-project autodetect (Anchor.toml shadows Quasar.toml) may bite Quasar-only users with a stale \`Anchor.toml\` left over from migration. \`--quasar\` is the escape hatch, so fine — but worth a thought for v0.4.

Both nits are pure documentation; behavior is correct and CI is green. Approving once they're addressed.

saicharanpogul:
Thanks for the careful review. Pushed [78dab3b](https://github.com/QEDGen/solana-skills/pull/20/commits/78dab3b) addressing all three:

**Nit 1 — parent-walk autodetect.** Went with the implementation rather than the docstring trim. `Framework::detect_from_cwd` now walks `std::env::current_dir()?.ancestors()` and picks the framework at the first level with either marker. Mixed (both at same level) still picks Anchor; mirrors how Cargo locates a workspace root, so devs running `qedgen` from a nested dir get the right autodetect.

**Nit 2 — test comment.** You were right that the comment was wrong, but the *actual* reason isn't the discriminator shape — both frameworks accept `[1,2,3,4,5,6,7,8]` as a `Vec<u8>` (Quasar's discriminator field is `Vec<u8>`, no length constraint). Verified empirically:

```
$ qedgen readiness --idl /tmp/anchor_idl_test.json --quasar
Error: ... Quasar IDL JSON: missing field `address` at line 16 column 1
```

It's the **top-level `address` field** — required in Quasar, nested under `metadata` in Anchor. Comment now reflects that. The test still proves what it needs to (dispatch reaches the Quasar parser).

**Nit 3 — Cargo.toml.** Alias-list block comment now names all three paths.

**Mixed-project autodetect (lower-priority you flagged).** Agree it's worth a thought for v0.4. The parent-walk doesn't change that semantic — same level with both markers still picks Anchor. `--quasar` remains the escape hatch. If you want, I can open a follow-up issue to track it on the qedgen side; the same question applies on ratchet's CLI too.

Test counts unchanged (442 total, 11 in `ratchet::tests`). fmt + clippy + drift all clean.

PR #44, opened 2026-05-15
feat: add pinocchio support for qedgen

AvhiMaz:
Adds Pinocchio program support to the adapt command. A new `pinocchio_adapt.rs` 
module auto-detects Pinocchio projects via `Cargo.toml`, parses dispatch arms and 
account layouts using regex, and emits a `.qedspec` scaffold.

### DIFF
--- a/crates/qedgen/src/main.rs
+++ b/crates/qedgen/src/main.rs
@@ -2,6 +2,7 @@ mod anchor_adapt;
 mod anchor_check;
 mod anchor_project;
 mod anchor_resolver;
+mod pinocchio_adapt;
 mod api;
 mod aristotle;
 mod asm2lean;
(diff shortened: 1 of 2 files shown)


abishekk92:
Thanks for putting this together — 438 lines, unit tests, and you call back into `chumsky_adapter::parse_str` to validate the emitted spec, which is the right reflex. Build is clean, the 10 unit tests pass. I want to surface what happens when this meets real Pinocchio code before we merge, because I think the shape needs a rethink rather than a polish.

## Tested on three real programs + one synthetic best-case

| Program | Result | Why it failed |
|---|---|---|
| [sontuphan/pinocchio-bulksender](https://github.com/sontuphan/pinocchio-bulksender) | error | dispatch lives in `src/processor/mod.rs`, not `lib.rs`; the discriminator is destructured *above* the match (`match discriminator { 0 => ... }`), so the `Some((N, rest)) =>` regex finds nothing |
| [Manudasari265/magicblock-delegate-pinocchio](https://github.com/Manudasari265/magicblock-delegate-pinocchio) | error | uses `match DelegationProgram::try_from(data)?` — match-on-enum, not `Some((N, ...))` |
| [NVN404/Pinocchio-Escrow](https://github.com/NVN404/Pinocchio-Escrow) (canonical Blueshift tutorial) | error | dispatch shape is right, but discriminator is `Make::DISCRIMINATOR` (a const), not a numeric literal — regex requires `(\d+)` |
| synthetic program built to match every assumption | parses, but mostly TODOs (see below) | — |

So the adapter works only on programs that simultaneously:
- put dispatch in `src/entrypoint.rs` or `src/lib.rs`
- match directly on `data.split_first()` with inline `Some((N, rest)) => fn_name(...)`
- use numeric-literal discriminators (not `Foo::DISCRIMINATOR`)
- have freestanding `fn process_xxx(...)` handlers (not `T::try_from(...)` builders)
- destructure accounts as `let [...] = accounts` (not `accounts.iter().next()`)
- define instruction data as `#[repr(C, packed)] struct` (not `#[derive(BorshDeserialize)]`)
- put errors in `src/error.rs` (singular)

That's not idiomatic Pinocchio — that's one specific tutorial layout.

## Even in the best case, output gaps

On the synthetic program that hits every assumption:
- accounts go into a `// accounts: x, y, z` *comment*, not an `accounts { ... }` block (so downstream lints/checks can't see them)
- every handler is `State.Init -> State.Init` — no transition inference
- underscored-but-real accounts (`_token_program`) are filtered out by `!name.starts_with('_')` in `extract_accounts`, but those are usually real accounts that just aren't bound in the body
- the rendered comment hardcodes `source: src/instructions/{name}.rs` regardless of which of the four candidate paths `parse_handler` actually found
- `is_pinocchio_project` falls back to substring match on parse failure, so a comment like `// pinocchio looks neat` in `Cargo.toml` would auto-route

## Architectural fit

The bigger concern: this is a regex-based parser of Rust source. QEDGen's general direction (see `feedback_agent_lsp_substrate.md` and how the audit subagent is shaping up) is *not* to build syn/AST scanners — instead, emit structured prompts and let the agent's rust-analyzer do the impl reading. `anchor_adapt.rs` gets away with heavier parsing because Anchor has a stable surface (`#[program]`, IDL). Pinocchio deliberately has no IDL and no canonical surface — every program freelances dispatch — which is exactly the case where pattern matching breaks first.

Also: no updates to README, SKILL.md, `references/cli.md`, no fixture in `examples/`. Our pre-release checklist would flag those.

## Suggested redirect

I think the win is real but the shape should flip. Concretely, what I'd want to see:

1. **No regex over Rust source.** Walk the `entrypoint!()` to locate the entrypoint function, list `src/` files, and write a structured prompt (`.qed/adapt/pinocchio-prompt.md`) describing what to fill in — the way the audit subagent does. The agent has rust-analyzer and can read impls without us re-implementing a Rust parser.
2. **No auto-detect.** Require an explicit `--pinocchio` flag (or, better, fold it into the prompt-emitting subagent path). The current auto-detect makes broken output look like first-class support.
3. **Real fixture.** One committed `examples/pinocchio-*` so codegen changes can't silently break this path.

If you want to keep the regex approach for a narrower v1, it'd need to (a) be opt-in, (b) document the exact shape it supports, (c) include at least one real-world fixture proving it survives.

Happy to chat through the prompt-emitting design if you want to take another swing — it's a much smaller diff and lands on the right side of the architecture line.

abishekk92:
Thanks for sending this, @AvhiMaz — and apologies for the timing mismatch on closing it.

Pinocchio support just shipped in [v2.19.0](https://github.com/QEDGen/solana-skills/releases/tag/v2.19.0) but along a different path than extending `qedgen adapt`. The v2.19 design routes Pinocchio (and Anchor / Native / qedgen-codegen) through a new scaffold-to-spec interview pipeline rather than a single-pass regex-to-spec emit:

1. `qedgen probe --program <root>` walks the program, detects the runtime, enumerates audit sites + handlers (Pinocchio path lives in `crates/qedgen/src/pinocchio_extractor.rs` and `crates/qedgen/src/pinocchio_probe.rs`).
2. `qedgen probe --emit-spec-candidates --audit-dir <path>` clusters findings into candidate spec clauses (`crates/qedgen/src/cluster.rs`) and writes an `interview.md` for the user.
3. `qedgen ratify --audit-dir <path>` merges the user's `accept` / `narrow` / `reject` / `bug` decisions into the final `.qedspec`, plus a rejected-clauses log in `.qed/plan/scoping.md` and bug-flagged finding files in `.qed/findings/`.

Why this shape instead of extending `qedgen adapt`:

- The same interview frontend works across all four supported runtimes, so adding a new framework reduces to "write an extractor", not "write an extractor + adapt code path + lint registration".
- The cluster + ratify step lets users surface *intentional* exclusions as `[x] reject` with a rationale, which we then capture as scoping notes. Single-pass regex emit can't distinguish "missing precondition because the spec author didn't model it" from "missing precondition because it's a real bug".
- For Pinocchio specifically, the bigger win in v2.19 was the audit-site catalogue (`pinocchio_probe.rs`) — `_unchecked` loads, `bytemuck` calls, raw-ptr casts, `set_lamports` / `set_amount` arithmetic, indexed-account access, etc., with Mollusk + Miri repro prompts attached. That sits upstream of the spec scaffold and would have been awkward to share between two parallel adapt paths.

The dispatch-arm and account-layout parsing logic in your PR overlaps `pinocchio_extractor.rs` in spirit — if you want to keep contributing, the highest-leverage spots right now are improving extractor coverage on dispatch shapes the current regex misses, and tightening cluster categorization in `cluster.rs`. Happy to chat on issues if you're interested.

Going to close this one — not because the work isn't sound, but because the design space moved between submission and review. Really do appreciate the contribution.

Issue #42, opened 2026-05-02
feat: support conditional effects (if/match) in handler effect blocks


saicharanpogul:
# Feature request — conditional effects (`if … then …`) in handler `effect` blocks

## Affected versions
- v2.15.1 (current — feature gap, not a bug)

## Summary
qedgen handler `effect` blocks support unconditional updates (`field += amount`, `field := value`) but no conditional form. This forces specs to either:
- Coarsely approximate handlers whose runtime effect depends on a parameter (e.g., `match` over a discriminant), losing verification precision; or
- Split a single program function into N spec handlers, even though the program runs as one instruction (breaks 1:1 spec/program mapping for drift detection).

A conditional effect form would let a single spec handler cleanly model a single program function whose behavior branches internally.

## Concrete pattern this blocks

A common Solana / Anchor pattern: one instruction that handles multiple cases via a discriminant parameter. Example — a fee-collection instruction:

```rust
pub fn collect_fees(ctx: Context<CollectFees>, fee_type: FeeType) -> Result<()> {
    match fee_type {
        FeeType::A => ctx.accounts.state.fees_a_withdrawn += amount_a,
        FeeType::B => ctx.accounts.state.fees_b_withdrawn += amount_b,
        FeeType::C => ctx.accounts.state.fees_c_accumulated = 0,
        FeeType::D => ctx.accounts.state.fees_d_accumulated = 0,
    }
    Ok(())
}
```

Today the spec must pick one effect target:

```
handler collect_fees (caller : U8) (fee_type : U8) : State -> State {
  ...
  effect { fees_a_withdrawn += amount }    // wrong for B, C, D
}
```

…and any property over the other counters becomes either vacuous or unverifiable for this handler.

## Proposed surface

```
handler collect_fees (caller : U8) (fee_type : U8) (amount : U64) : State -> State {
  ...
  effect {
    if fee_type == 0 then fees_a_withdrawn += amount
    else if fee_type == 1 then fees_b_withdrawn += amount
    else if fee_type == 2 then fees_c_accumulated := 0
    else fees_d_accumulated := 0
  }
}
```

Any of the standard variants would work — `if/else if/else`, `match`, or guarded effect blocks (`when fee_type == 0 effect { … }`). Match on a U8 discriminant is the most common case.

## Workaround in current versions
- Restrict the spec handler's effect to one branch and `preserved_by [list-without-this-handler]` the affected properties — every other handler is still verified, the branch-dependent handler is honestly excluded.
[...]

PR #45, opened 2026-05-15
fix(proptest_gen + adapter): make standalone proptest compile for larger brownfield specs

0xharp:
> Tried qedgen on your nudge from [this thread](https://x.com/shek_dev/status/2054419082439704773) — wanted to try it on our [tendr.bid](https://github.com/0xharp/tender) Anchor program, which has a decent number of instructions. These are the codegen edges I hit while integrating, with fixes.

## TL;DR

Six small, isolated codegen fixes in `proptest_gen`, `chumsky_adapter`, and `rust_codegen_util`, plus one regen commit catching up the affected bundled examples. Each fix has a reproducer, ships behind a usage gate where applicable, and is verified against the full `ci.yml` gate sequence.

## What

Surfaced while integrating qedgen with [tendr.bid](https://github.com/0xharp/tender) — a brownfield Anchor program with a flat-state `.qedspec`. Each commit is independent (the regen commit at the end depends on the prior six), with a focused message explaining the bug + the fix + the surface where it shows up.

| # | Commit | What it fixes |
| - | --- | --- |
| 1 | `fix(chumsky_adapter): widen both sides when coercing Nat to Int` | Code path: spec compares an `I*` field against a Nat literal (`>= 0`). `render_rust_binary_with_coercion` cast only the Nat side to `i128`, leaving an `i64 >= i128` mismatch (E0308). Now widens both sides symmetrically. |
| 2 | `fix(rust_codegen_util): bind state-field idents via target-aware binder in resolve_value` | Effect-RHS rendering strips the `state.` prefix upstream (so each backend can apply its own state binder) but `resolve_value` then returned the bare ident, which doesn't bind in the target's body scope. Adds a `state_binder: Option<&str>` parameter so each caller passes its own (`Some("s.")` for proptest, `Some(&format!("self.{}.", acct))` for Anchor/Quasar handlers). `None` preserves legacy bare-identifier behavior. |
| 3 | `fix(proptest_gen): type-aware seed-state init for arrays + lifecycle status` | Three issues at the same code path: `Map[N] T` arrays got `field: 0` (E0308); record-typed array elements got `field: [0; N]` against `[Account; N]` (E0308); the `status:` lifecycle discriminator was hardcoded to `Status::Uninitialized`, missing for specs whose first state isn't named that (e.g. percolator: `Active`, E0599). Now uses a recursive `default_value_for_field` helper that emits `[<inner>; N]`, `<RecordName>::default()`, `0`, or `None` (skip + let rustc surface the missing field), and reads `initial_lifecycle_state(spec)` for the discriminator. |
| 4 | `fix(proptest_gen): inline mul_div helpers when standalone proptest needs them` | Brownfield workflow runs `--proptest` without `--all`, so `src/math.rs` isn't generated and `mul_div_floor_u128` calls don't resolve. Inlines the helpers — gated on `codegen::guards_use_math_helpers` (now `pub(crate)`) so specs that don't use them don't carry duplicate definitions. The same helper was extended to also probe `let_bindings.rust_expr` (specs that use mul_div in handler-level let bindings, like fee math, were previously missed). |
| 5 | `fix(proptest_gen): chunk arb_op tuples for handlers with >12 args` | proptest's `Strategy` impls cap at tuple arity 12; >12-arg handlers overflow with E0599. Chunks strategies into ≤12-arity sub-tuples with a nested destructuring pattern; ≤12-arg handlers keep their existing flat shape. |
| 6 | `fix(proptest_gen): emit recursion_limit override only when needed` | rustc's default `recursion_limit = 128` overflows on `arb_state()` for ≳40-field States. Emits `#![recursion_limit = "512"]` only when the spec's total mutable field count exceeds 32 — bundled examples (escrow ≈ 3, lending ≈ 6, multisig ≈ 5, percolator ≈ 4) keep the default. 512 = 2× rustc's suggested 256, empirically validated against a 99-field flat State. |
| 7 | `chore(examples): regen bundled codegen for proptest_gen + adapter fixes` | Catches up the six affected bundled-example artifacts so `qedgen check --regen-drift` stays clean. Per-file rationale in commit body. |

After these, our `.qedspec` produces a proptest harness that compiles and runs cleanly, and the same patches make the kani auto-gen compile too (both backends share the underlying renderers, so the fixes propagate).

## Context

We were integrating qedgen (v2.18) with [tendr.bid](https://github.com/0xharp/tender) — a sealed-bid procurement program on Solana with ~28 handlers, a fairly large flat State after canonical-pattern flattening, and `mul_div_floor` for fee math. The combination of all three (large flat State + signed time arithmetic + brownfield `--proptest` without `--all`) is what surfaced these specific edges.

Worth flagging in case useful: the multisig proptest output has separate issues (`Map[N] Pubkey` → `[Address; N]` placeholder, auth-alias unbinding) that felt more like design questions than clear bugs, so I've kept them out of scope here — happy to file a discussion issue if you'd like. The CI smoke gate runs `cargo test` against escrow's proptest harness only (the other examples are scaffold-compile only), so these particular paths just hadn't been exercised before.

## Verification

Validated end-to-end on a fork-CI run that exercises the full `ci.yml` gate sequence — link will be added once force-push triggers it. Specifically:

- `cargo fmt --check` — clean
- `cargo clippy -- -D warnings` — clean
- `cargo test` — full workspace: 568 passed, 0 failed, 8 ignored (the 4 codegen_smoke and 4 doctests are intentionally `#[ignore]` and run separately in the next gate step)
- `bash scripts/check-version-consistency.sh` — clean
- `bash scripts/check-readme-drift.sh` — clean
- `cargo run -p qedgen-solana-skills -- check --regen-drift` — clean
- `cargo test --test codegen_smoke -- --ignored` — 4 passed, 0 failed

No new unit tests added in this PR — relied on the existing workspace test suite (568 tests across all crates, ≈543 in `qedgen` itself) as integration check, plus the regen commit (#7) acts as a downstream-codegen integration gate. Happy to add focused tests for any of the new code paths if useful.

Local repro note: with `core.autocrlf = true` (common on WSL/Windows), the bundled snapshot fixtures and regenerated outputs disagree on line endings, which makes a few snapshot tests + the percolator smoke fail locally even though they're green on Ubuntu CI. Setting `core.autocrlf = false` for the workspace makes the local run match CI.

## Things deliberately NOT included

- **Pubkey codegen.** Touched on this area too (the `Map[N] Pubkey → [Address; N]` pattern needs an `Address` definition; `auth <name>` aliases would need binding in handler-fn bodies; `mutable_fields` filters bare `Pubkey` so properties referring to Pubkey state fields can't currently find the field). These felt like design calls more than bugs, so I'd rather raise them as a discussion issue than touch them here.
- **A patch for narrowing `mul_div_floor_u128` results to `u64`.** We carry it locally because our spec uses u64 amounts throughout, but it'd be the wrong fix for percolator-style specs that genuinely want u128 amounts — needs a real type-inference pass at the let-binding site, out of scope for this batch.
- **Broader bundled-example regen** beyond the six artifacts directly affected by these codegen changes is intentionally out of scope here, to keep the diff focused. Happy to send a follow-up regen sweep separately if useful.

Would love your read on whether the framing of these fixes makes sense — happy to split, squash, or rework any of them. If any of this doesn't match the design intent, please feel free to push back or close the PR; happy to learn what we got wrong. Thanks for building qedgen, the spec-driven flow has been a pleasure to work with end-to-end.

### DIFF
--- a/crates/qedgen/src/chumsky_adapter.rs
+++ b/crates/qedgen/src/chumsky_adapter.rs
@@ -973,9 +973,13 @@ fn render_rust_binary_with_coercion(
     let rk = rust_infer_kind(opts.env, &rhs.node);
     let l = expr_to_rust(&lhs.node, ctx, consts, opts);
     let r = expr_to_rust(&rhs.node, ctx, consts, opts);
+    // When widening Nat → Int we must cast BOTH sides to the same wide
+    // type. Pre-fix only the Nat side was cast, leaving comparisons like
+    // `i64 >= i128` that don't typecheck. Symmetric widening to i128 keeps
+    // operands aligned without losing precision on either side.
     match (lk, rk) {
-        (Kind::Nat, Kind::Int) => (format!("(({}) as i128)", l), r),
-        (Kind::Int, Kind::Nat) => (l, format!("(({}) as i128)", r)),
+        (Kind::Nat, Kind::Int) => (format!("(({}) as i128)", l), format!("(({}) as i128)", r)),
+        (Kind::Int, Kind::Nat) => (format!("(({}) as i128)", l), format!("(({}) as i128)", r)),
         _ => (l, r),
     }
 }
--- a/crates/qedgen/src/codegen.rs
+++ b/crates/qedgen/src/codegen.rs
@@ -2311,7 +2311,7 @@ fn precompute_body_hash(scaffold_source: &str) -> Option<String> {
 /// fixed-point helpers in `src/math.rs`. Used to gate the `use crate::math::*;`
 /// import in `guards.rs` so legacy programs whose user-owned `lib.rs` doesn't
 /// declare `pub mod math;` keep compiling.
-fn guards_use_math_helpers(spec: &ParsedSpec) -> bool {
+pub(crate) fn guards_use_math_helpers(spec: &ParsedSpec) -> bool {
     let mut any = false;
     let probe = |s: &str| s.contains("mul_div_floor_u128") || s.contains("mul_div_ceil_u128");
     for h in &spec.handlers {
@@ -2322,6 +2322,14 @@ fn guards_use_math_helpers(spec: &ParsedSpec) -> bool {
             any = true;
         }
         if h.ensures.iter().any(|e| probe(&e.rust_expr)) {
+            any = true;
+        }
+        // Handler-level `let bindings: (lean_expr, rust_expr)` also lower to
+        // `let X = mul_div_floor_u128(...)` in the emitted Rust handler body.
+        // Without this, specs that compute fee math via a `let` (a common
+        // pattern for splitting amounts before the effect block) wouldn't
+        // pick up the math.rs import / inline helpers.
+        if h.let_bindings.iter().any(|(_, _, r)| probe(r)) {
             any = true;
         }
     }
--- a/crates/qedgen/src/codegen.rs
+++ b/crates/qedgen/src/codegen.rs
@@ -1858,8 +1858,12 @@ fn mechanize_effect(
         return None;
     }
 
-    let rhs = crate::rust_codegen_util::resolve_value(value, handler, spec);
+    // Anchor / Quasar handler bodies bind state as `self.<acct>.<field>`,
+    // so a bare state-field RHS (e.g. `bid_buyer := state.rfp_buyer` after
+    // upstream strips `state.`) needs to resolve to `self.<acct>.rfp_buyer`.
     let acct = &state_acct.name;
+    let acct_binder = format!("self.{}.", acct);
+    let rhs = crate::rust_codegen_util::resolve_value(value, handler, spec, Some(&acct_binder));
     // Cast index expressions in the LHS path to `usize`. `render_effect`
     // emits the field as `voted[member_index]` (Lean-friendly); on the
     // Rust side, indexing `[u8; N]` with `u8`/`u16`/Fin fails — Rust
--- a/crates/qedgen/src/proptest_gen.rs
+++ b/crates/qedgen/src/proptest_gen.rs
@@ -404,6 +404,36 @@ pub fn generate(spec_path: &Path, output_path: &Path) -> Result<()> {
     );
     out.push_str("use proptest::prelude::*;\n\n");
 
+    // ── Math helpers ─────────────────────────────────────────────────────
+    // The brownfield workflow runs `qedgen codegen --proptest` against an
+    // existing program crate, without `--all`. In that case `src/math.rs`
+    // is never generated, so any `mul_div_floor_u128` / `mul_div_ceil_u128`
+    // calls emitted by chumsky_adapter::expr_to_rust have no definition in
+    // scope. Inline the helpers ONLY when the spec actually uses them —
+    // otherwise we'd ship two sources of truth for every spec that
+    // doesn't need them, with the silent-divergence risk that implies for
+    // any future change to the canonical helper.
+    //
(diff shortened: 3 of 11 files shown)


abishekk92:
Rebased onto `main` post-v2.20.0. One conflict in `rust_codegen_util.rs` — your `state_binder: Some("s.")` fix lifted into the shared `emit_one_effect` helper so the new v2.20 `match`-block path (S1.2 conditional effects, #42) also benefits. CI clean locally (fmt, clippy `-D warnings`, full test suite).

Worth noting: these six fixes compose cleanly with v2.20. PR #45 is compile-shape (Nat/Int coercion, state binding, large-state recursion, mul_div inlining, tuple-arity chunking, conditional recursion_limit); v2.20 is correctness-shape (forall harness no longer stubs to `true`, `match` in effect blocks, Pubkey P6 lint). The Pubkey-codegen item you flagged as out-of-scope is now visible behind the v2.20 P6 lint — handler-parameter workaround documented in `docs/limitations.md#pubkey-state-fields`; codegen-side fix (`Bytes32` spec type or auto-lowering) tracking for v2.21.

Thanks for the careful split and the per-commit reproducers — made the rebase mechanical.

abishekk92:
Quick status update after v2.20 was merged + tagged.

**Main is now green** — CI fixes that landed post-rebase: `clippy::collapsible_match` on `shank_probe.rs` (rust 1.95 default-profile lint), `package.json` version bump to match `Cargo.toml`, and a `check --regen-drift` catch-up commit for bundled examples whose codegen output shifted under S1.1's `forall` harness lowering.

**PR #45 is rebased onto current main** (`d117f12`), conflicts already resolved (one in `rust_codegen_util.rs` — `state_binder` lifted into `emit_one_effect`). However, **`Check example codegen drift` will fail** on this branch because your `chore(examples): regen bundled codegen` commit captured codegen output from before v2.20 landed. The substantive output shape changes (S1.1 forall lowering, S2.2 per-handler categories, etc.) make the committed regen stale.

Two paths forward, your call:

1. **Squash a fresh regen into the existing regen commit.** Pull the rebased branch (`git fetch && git checkout proptest-codegen-fixes && git reset --hard origin/proptest-codegen-fixes`), run `cargo build --release && target/release/qedgen check --regen-drift` to identify drifted files, regen per-example with the right `--target`, amend into commit `d117f12`, force-push.
2. **I do the regen catch-up as a maintainer follow-up commit on top of the PR.** Slightly noisier history but unblocks merge immediately.

Either way the 6 fix commits themselves stand on their own — the regen drift is a downstream cost of rebasing across a major minor.

0xharp:
Thanks @abishekk92, appreciate the rebase and the post-v2.20 catch-up commit on your side. Especially nice to see state_binder: Some("s.") lifted into the shared emit_one_effect helper so the v2.20 match-block path picks it up too — better placement than the per-call sites in the original PR.

The framing in your first comment (compile-shape vs correctness-shape) matches how we hit them, the six fixes here were helpful for a brownfield spec from generating standalone proptest / Kani harnesses that compile. v2.20's correctness-shape work (forall lowering, match in effects) is downstream of that, couldn't have benefited from it without the compile path working first. Glad they compose cleanly.

On the Pubkey item and tracking for v2.21 (Bytes32 / auto-lowering): just pulled v2.21.0 and trying the new Pubkey → [u8; 32] state-field lowering against our spec now — happy to follow up with whether the existing U64-Pubkey workaround in our spec drops cleanly, or if anything else surfaces that's useful as signal for the v2.21+ direction.

Thanks again for the speed on the merge and for taking on the regen catch-up rather than bouncing it back. Made shipping this end-to-end much easier on our side.

PR #46, opened 2026-05-19
fix(kani): standalone-harness compile-shape fixes (binder + mul_div inline + array seed init)

0xharp:
Continued from where PR #45 left off — taking v2.21.0 for a spin on our [tendr.bid](https://github.com/0xharp/tender) Anchor program. Three more codegen edges surfaced on the standalone kani harness path that the bundled examples don't quite exercise. Each is small and isolated; full validation evidence below. Thank you for the speed on the v2.20 and v2.21 releases, by the way — it's been a great experience tracking the cutting edge.

## The fixes

**1. `fix(kani): resolve effect-conformance RHS via target-aware binder`**

v2.21's per-handler effect-conformance harness (the per-field `verify_X_effect_Y()` fns) emits the RHS as a bare identifier, so any spec whose effect block copies a state field into another state field fails to compile:

```
error[E0425]: cannot find value `now` in this scope
   --> tests/kani.rs:N:M
    |
  N |         assert!(s.rfp_created_at == now, "...");
    |                                    ^^^
```

The snapshot binding `pre_now` is already declared earlier in the same fn — fix is one `resolve_value(.., Some("pre_"))` call ahead of the match arms. `resolve_value` is identity for params and inlines constants, so any effect whose RHS isn't a state-field name is unchanged.

**2. `fix(kani): inline mul_div helpers when standalone kani harness needs them`**

The standalone `tests/kani.rs` doesn't `pub use crate::math::*`, so any spec calling `mul_div_floor` from an effect's let-binding fails to compile with `error[E0425]: cannot find function 'mul_div_floor_u128'`. Inline the canonical helper bodies right after `#![cfg(kani)]`, gated by `guards_use_math_helpers` so specs that don't use mul_div get a zero-byte change.

**3. `fix(kani): type-aware seed-state init via shared default_value_for_field`**

`emit_state_init_zeroed` literaled every field to `0` regardless of type. With v2.21's S3 Pubkey lowering and any `Map[N] T` field, that produces:

```
error[E0308]: mismatched types
   --> tests/kani.rs:N:M
    |
  N |         rfp_milestone_amounts: 0,
    |                                ^ expected `[u64; 8]`, found integer
```

Reuse the existing `default_value_for_field` helper (visibility bumped to `pub(crate)`) which already handles `Map[N] T`, records, Pubkey, sums, and primitives. The three call sites in `emit_kani_account_section` thread `spec` through.

## Chore commit

`chore(examples): regen bundled kani harnesses for kani.rs fixes` catches up the five affected `tests/kani.rs` files (escrow, lending, multisig + multisig/programs, percolator) so `qedgen check --regen-drift` stays clean on the kani path. Scope is intentionally tight — only the five files these fixes change. Other v2.21 drift under `src/integration_tests.rs` / `Cargo.toml` / `qed.lock` is left for separate maintainer cleanup if needed.

## Validation

- `cargo test -p qedgen-solana-skills --bin qedgen`: 688 passed, 0 failed, 1 ignored
- `cargo clippy --release -p qedgen-solana-skills -- -D warnings`: clean
- `cargo fmt --check`: clean
- Bundled example byte-identity vs stock v2.21.0 binary on all 5 examples: every diff line falls in one of the three known fix categories (mul_div helper insert, Pubkey/array defaults, pre_ binder rewrite) — verified with diff
- `qedgen check --regen-drift` on the kani path: clean
- `bash scripts/check-lake-build.sh --strict` (lake-build workflow, manual dispatch on this branch): success in 4m22s — Lean side unaffected

## Heads-up on pre-existing CI

For visibility (so it doesn't read as a regression from this PR): upstream main's CI is also currently red on the v2.21.0 release commit ([run `26032919566`](https://github.com/QEDGen/solana-skills/actions/runs/26032919566)) with the same 5 failing tests this PR's CI will also show:

- `escrow_codegen_is_deterministic`
- `escrow_split_multi_spec_is_deterministic`
- `lending_codegen_is_deterministic`
- `multisig_codegen_is_deterministic`
- `percolator_codegen_is_deterministic`

Looks like `crates/qedgen/tests/codegen_determinism.rs` calls `qedgen codegen --all` which gates on Kani install, but the `test` job in `.github/workflows/ci.yml` doesn't install Kani before running tests. Happy to send a small follow-up PR adding a `cargo install --locked kani-verifier && cargo kani setup` step (or whatever shape you'd prefer) if useful — totally fine if you'd rather handle it separately.

Thanks again — please let me know if anything needs reshaping. Happy to correct my understanding if I've inadvertently shipped something wrong here.

### DIFF
--- a/crates/qedgen/src/kani.rs
+++ b/crates/qedgen/src/kani.rs
@@ -921,24 +921,43 @@ fn emit_kani_account_section(
                     .collect();
                 out.push_str(&format!("    if {}(&mut s{}) {{\n", op.name, args));
 
-                // Assert THIS field's effect only
+                // Assert THIS field's effect only.
+                //
+                // The effect-conformance harness snapshots every mutable
+                // field as `pre_<fname> = s.<fname>` BEFORE calling the
+                // transition. The post-condition RHS — `value` here — comes
+                // from the spec's effect block (e.g. `:= state.now`), with
+                // the `state.` prefix already stripped by the upstream
+                // chumsky_adapter so each backend can apply its own binder.
+                //
+                // Without binder resolution the emitted assertion reads
+                // `assert!(s.X == now)` — bare `now` is undefined in scope
+                // and the harness fails to compile with
+                // `error[E0425]: cannot find value 'now' in this scope`.
+                //
+                // `resolve_value` is identity for handler params and inlines
+                // constants; the `Some("pre_")` binder is applied only when
+                // `value` is a state field name. Same shape as the
+                // analogous fix in `rust_codegen_util::emit_one_effect` for
+                // the transition-fn emission target (`Some("s.")`).
+                let resolved = rust_codegen_util::resolve_value(value, op, spec, Some("pre_"));
                 match op_kind.as_str() {
                     "set" => {
                         out.push_str(&format!(
                             "        assert!(s.{} == {}, \"{} must equal {}\");\n",
-                            field, value, field, value
+                            field, resolved, field, resolved
                         ));
                     }
                     "add" => {
                         out.push_str(&format!(
                             "        assert!(s.{} == pre_{}.wrapping_add({}), \"{} must increment by {}\");\n",
-                            field, field, value, field, value
+                            field, field, resolved, field, resolved
                         ));
                     }
                     "sub" => {
                         out.push_str(&format!(
                             "        assert!(s.{} == pre_{}.wrapping_sub({}), \"{} must decrement by {}\");\n",
-                            field, field, value, field, value
+                            field, field, resolved, field, resolved
                         ));
                     }
                     _ => {}
--- a/crates/qedgen/src/kani.rs
+++ b/crates/qedgen/src/kani.rs
@@ -41,14 +41,35 @@ fn emit_state_init_symbolic(
 /// `status` field set to the section's initial lifecycle state. Used by init-
 /// handler harnesses (effect/preservation), where the pre-state is the
 /// canonical "before initialization" state.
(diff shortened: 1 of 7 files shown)


abishekk92:
Thanks for the careful test-drive and tight scope, @0xharp — exactly the kind of bench-driven feedback that lands cleanly. All three fixes plus the example regen are folded into `feat/v2.22` as the merge commit `344c38a` ([branch](https://github.com/QEDGen/solana-skills/tree/feat/v2.22)). The four commits will reach `main` when v2.22 ships.

Validation on top of the v2.22 branch (10 commits ahead of v2.21.0):
- `cargo build --bin qedgen`: clean
- `cargo test --bin qedgen`: 721 passed
- `cargo fmt --check`: clean
- `cargo clippy -- -D warnings`: clean

On the CI-redness heads-up — yes, the `codegen_determinism` tests need Kani installed and the `test` job currently doesn't `cargo install kani-verifier`. Will sort that separately.

0xharp:
awesome, thanks @abishekk92 🙏 . excited for v2.22.

abishekk92:
All four commits landed on main via v2.22.0 release ([PR #47](https://github.com/QEDGen/solana-skills/pull/47), rebased so commit hashes differ but tree-content is identical). Thanks again @0xharp — these fixes are now in [v2.22.0](https://github.com/QEDGen/solana-skills/releases/tag/v2.22.0).

0xharp:
🙏 thank you @abishekk92, pinning to v2.22 now.

Issue #79, opened 2026-06-02
[qedgen 2.33.0] codegen lowering: Lean effect-RHS field ref + proptest chained-subtraction


abishekk92:
## What happened

Two **codegen lowering** correctness bugs in `qedgen 2.33.0`, both reproducible from one tiny spec, and notably **inconsistent between the proptest and Lean backends**. Found while writing a small spec with a per-period budget invariant and proving it in Lean (Phase 2). Neither surfaces as a `qedgen` error — codegen succeeds and the defect appears downstream (`lake build` / `cargo test`).

A self-contained repro spec (`budget.qedspec`) is excerpted below; both bugs fire from it.

---

## Bug A — Lean: effect-RHS field reference loses the `s.` qualifier

A handler effect that assigns one state field from another state field lowers the RHS to an **unqualified identifier** instead of `s.<field>`. It only compiles when a handler *parameter* of the same name happens to be in scope.

Spec:
```
handler reset : State.Active -> State.Active {
  permissionless
  accounts { caller : signer }
  effect { used := 0
           reserved := state.cap }     // also reproduces with bare `reserved := cap`
}
```
Generated `Spec.lean`:
```lean
def resetTransition (s : State) (signer : Pubkey) : Option State :=
  if s.status = .Active then
    some { s with used := 0, reserved := cap, status := .Active }   -- ❌ unknown identifier `cap`
  else none
```
`lake build` fails (`unknown identifier 'cap'`). Expected `reserved := s.cap`.

Asymmetry: `open (c : U64)` emits `cap := c` **correctly**, because `c` resolves to the handler parameter. So the lowering resolves RHS names against handler params but not against state fields — and `state.cap` is stripped to bare `cap` rather than `s.cap`.

---

## Bug B — proptest: chained subtraction mis-lowered (wrong associativity + injected `wrapping_sub`)

A guard with chained subtraction lowers incorrectly in the **proptest** backend, while the **Lean** backend lowers the same expression correctly — so the two backends disagree.

Spec:
```
handler use_budget (amount : U64) : State.Active -> State.Active {
  permissionless
  accounts { caller : signer }
  requires amount <= state.cap - state.used - state.reserved else Exceeds
  effect { used += amount }
}
```
Generated **proptest** (wrong):
```rust
if !((amount <= s.cap - s.used.wrapping_sub(s.reserved))) { ... }
//                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #80, opened 2026-06-02
fix(qedgen): infer Pinocchio Kani impl profiles

candyflipline:
Adds a generic Pinocchio proof-profile layer for `--kani-impl` so the Kani emitter can consume parser-backed source facts and ABI-derived dispatcher tags, account order and roles, instruction packing, PDA derivations, token account bindings, repeated records, fixed bytes, and account layout facts without selecting custom code by program name. This also adds a richer generic Pinocchio proof-lab fixture whose accepted proof-completion harnesses call the real dispatcher, assert success, assert concrete post-state facts, and require Kani success-path cover evidence.

The paired-validator probe now ignores `#[cfg(kani)]` proof-model files/functions when comparing production validator domains, avoiding false positives from proof-only equality helpers while preserving the canonical paired-validator mismatch checks. Unsupported source shapes stay on conservative fallback paths, and the older string-based source extractor is used only when Rust parsing fails.

Validation run locally:

```sh
cargo fmt --manifest-path Cargo.toml --all
git diff --check
cargo build -p qedgen-solana-skills --bin qedgen
cargo test -p qedgen-solana-skills paired_validator -- --nocapture
cargo test -p qedgen-solana-skills pinocchio_kani_profile_diversity_fixture_generates_expected_proofs --test codegen_smoke -- --nocapture
cargo test -p qedgen-solana-skills kani_impl -- --nocapture
QEDGEN_RUN_CARGO_KANI_SMOKE=1 QEDGEN_KANI_HARNESS_TIMEOUT=300 cargo test -p qedgen-solana-skills generated_pinocchio_profile_diversity_kani_impl_proves_with_cargo_kani --test codegen_smoke -- --ignored --nocapture
```


abishekk92 (commented):
## Review — Pinocchio Kani impl profiles

Thorough pass across `pinocchio_profile.rs`, the `kani_impl.rs`/`kani_mir.rs` emitter rewrite, `cpi_substitute.rs`, `rust_codegen_util.rs`, the probes, and the new example/snapshots. **Overall this is well-engineered and genuinely program-agnostic** — I found no fixture/program-name special-casing; everything keys off Pinocchio conventions (`process_*`, `derive_*`, `require_token_account`, `find_program_address`, slice-destructure). The string scanners in `rust_codegen_util.rs` are paren/bracket/string-depth aware and unit-tested, the syn→regex fallback path degrades conservatively, and `cpi_substitute` + the probe false-positive suppressions are clean and tested.

No hard blockers. A few things I'd want addressed before merge:

### Major

1. **Silent wrong-layout: `bool`/`bytesN` scalars dropped from instruction-data packing.** `pinocchio_profile.rs:870-877` (`abi_field_to_profile_param`) returns `None` for any field that isn't an integer or `pubkey`, while the emitter packs params with **sequential `extend_from_slice`** (`kani_impl.rs` ~275-287/333-345), not by `field.start`/`field.end`. A record like `{ bool flag; u64 amount }` yields a harness that writes `amount` at byte 0 while the real `process_*` reads `flag` at 0, `amount` at 1 — a confidently-wrong harness with no TODO breadcrumb. This defeats the harness silently rather than failing conservatively. Please pad dropped fields with symbolic bytes, pack by absolute offset, or at minimum emit a `// TODO: unpacked field at offset N`. (Note: the existing `MEMO bytes8` test puts the dropped field *last*, so the drift never manifests — a test with a non-trailing bool/bytes field would catch this.)

2. **No snapshot fixture for the new Pinocchio impl harness.** The ~1500-line profile-driven Pinocchio emission (PDA seeds, token bindings, nested key derivations, instruction packing) is the heart of the PR but is covered only by in-module `#[test]`s. Every other Kani path has a checked-in `*.kani.rs` snapshot; without one here, regressions in the most complex new code won't be caught by `kani_snapshot.rs`. The `ptoken-transfer` fixture referenced at `kani_impl.rs:2826` looks like the intended candidate.

3. **New example is a test fixture in the showcase dir.** `examples/rust/kani-cpi-account-bindings/` contains only `config.qedspec` + `qed.toml`/`qed.lock`/`.qed/config.json` — no program code, no `formal_verification/`, no README — and is wired straight into `tests/kani_snapshot.rs` (`snapshot_kani_cpi_account_bindings`). The `stable_swap_large_guard` handler is synthetically contrived (9 `requires`, `lane < 32`, bps mul/div) to exercise the new rewrites. Per CLAUDE.md + the examples-vs-fixtures rule, `examples/` is showcase-only; synthetic snapshot inputs belong under `crates/qedgen/tests/fixtures/`. Relocate and repoint the snapshot.

### Minor / soundness edges

- **Multi-transfer token post-assertions can false-positive.** `emit_pinocchio_token_post_assertions` asserts each transfer independently; a handler that does A→B then B→C asserts B is both `pre_B + amt0` and `pre_B - amt1` → spurious counterexample. Single-transfer (all current fixtures) is fine — guard or document as single-transfer-only. Same class: self-transfer aliasing (`from == to`) isn't constrained; a `kani::assume(from_key != to_key)` would close it.
- **Duplicated syn/regex inference paths** (`pinocchio_profile.rs` ~1818-2489 mirror the syn twins) — ~600 lines that must stay in lockstep; the bool-drop bug above would need fixing in both. Worth considering whether an unparseable file should just yield an empty profile (the conservative fallback the rest of the design already relies on) instead of a second full implementation.
- **Unconditional progress `eprintln!` spam** on every `qedgen codegen --kani` (`kani_mir.rs:4398`, `render(progress=true)`) — no `--quiet`. Snapshots use `progress=false`; interactive users get ~15 lines. Gate behind a verbosity flag.
- Inferred account ordering is discarded wholesale on a single name mismatch (`kani_impl.rs:948-953`, exact `acct.name ==` against normalized names) — a `tokenProgram` vs `token_program` skew silently loses the entire reordering (the main value-add over spec order). Normalize both sides or add a debug breadcrumb.
- `account` schema rows with an unparseable index are silently dropped (`pinocchio_profile.rs:672-684`); non-PDA account keys are pinned to distinct concrete constants (inherent model limitation, matches prior behavior — flag only).

### Nits
- `_pinocchio_call_path` / possibly-unused `rust_string_escape` (`kani_impl.rs:1699`) — confirm dead code is intentional.
- Orphaned doc comment left above `anchor_lib_disambiguates_…` test in `codegen_mir.rs:1524-1526` after the duplicate-`#[test]` cleanup.

### Well done
- `cpi_substitute` partial-vs-total binder distinction, properly tested; PDA derivations use `kani::assume(pda.is_some())` rather than pinning to a constant (sound); `build_data_account` takes `kani::any()` and overwrites only fixed/discriminant bytes (correctly under-constrained); `verify.rs` `prepare_standalone_kani_crate` isolates the spec-model harness so unrelated scaffold compile errors no longer block Kani; both probe changes reduce false positives without adding advisory-only findings.

### Heads-up on merge order
This touches `rust_codegen_util.rs` and `lean_gen_mir.rs`, which #81 (the v2.33.1 issue-#79 fix) also changes. I checked: the regions are disjoint — #80's `rust_codegen_util.rs` additions sit between `translate_guard_to_rust` and `translate_property_to_rust` and refactor `emit_one_effect`, while #81 only touched `wrap_arith_expr`/`wrap_arithmetic_atom` (after `translate_property_to_rust`) and the `emit_handler_transition`/`build_guard_cond_parts` flat-state RHS in `lean_gen_mir.rs`. Should merge cleanly, but whichever lands second should re-run the codegen/kani/proptest snapshot suites since both regenerate fixtures.

abishekk92:
Pushed `00a8826` (from `fix/kani-vacuous-proofs`) onto this branch — two correctness fixes in the generated Kani harnesses, both of which the string-only snapshot gate accepted silently:

1. **Vacuous guard-rejection proofs past 8 guard terms** — the per-term split path assumed `!(term)` and then asserted `!(term)` back (`assume P; assert P`), never invoking the transition. Specs with >8 `requires` clauses silently lost guard-enforcement verification. The split harness now assumes the prefix terms true + the i-th violated, then asserts the handler actually rejects.
2. **`pubkey_eq` gated to scalar operands** so the helper isn't applied to non-scalar comparisons.

Also fixes the red unit test this PR shipped. Branch fast-forwards cleanly; PR is mergeable against current main (which now includes the #86 un-vendoring).

abishekk92:
Local review of `00a8826` (verified empirically — real `cargo kani` runs on the split harnesses, including a mutation teeth-test): **approve with fixes**, which are now pushed as `1c6f9b2`.

**Verified:** the >8-guard-term split fix is sound — every arm invokes the handler, the arms partition the violation domain exactly (first-violated-term decomposition), and deleting a guard from the transition turns the corresponding arm red. The unit test gates reintroduction of the `assume P; assert P` tautology.

**Fixed in `1c6f9b2`:**
1. `pinocchio_unchanged_ensures_fields` matched `post.X==pre.X` as a bare substring, so a conservation ensures (`state.X == old(state.X) + delta`) bogusly classified `X` as unchanged → guaranteed false `VERIFICATION FAILED` on a correct implementation. Match is now anchored (comparison must end the expression or be followed by a connective/closing paren); regression test covers the `pre + delta` shape in both orientations.
2. Guard-rejection harnesses (monolithic + split arms) now emit `kani::cover!(true, "guard-violation domain is satisfiable")` — an arm whose term is implied by its prefix (e.g. arm 9 of the bundled fixture, implied by terms 6+7) had UNSAT assumptions and reported SUCCESSFUL while proving nothing. The unsatisfied cover now fails loudly and flags the redundant guard term. Note: running `cargo kani` on the bundled fixture's arm 9 will now report this — that's the intended signal.

Snapshots + committed example harnesses regenerated; `check --regen-drift` clean; full suite green.

**On record for follow-up (not blocking):** the cargo-kani proof-completion e2e tests self-skip in CI (`QEDGEN_RUN_CARGO_KANI_SMOKE` gate) while reporting `ok`; the profile layer retains fixture-shaped constants (`lane_count→2`, `lane_id < 2`, the 7-param fee_normalization shape) that narrow the verified domain; ABI repeat packing assumes a u8 count byte immediately before the repeat region; and the "multisig harness now compiles" claim in the earlier commit message is inaccurate — its 12 compile errors are byte-identical at the merge-base (pre-existing, just not worsened).

Merge risk with current main is low: clean auto-merge, disjoint regions in shared files, merged tree verified green locally (987 unit tests + all four snapshot suites).

PR #132, opened 2026-06-24
feat(discharge): A2a — persist discharged artifacts into the project

abishekk92:
Slice A step **A2a** (`docs/design/qedsvm-discharge.md` §14). The qedgen-side plumbing that stops throwing the proof away.

## Before / after
`qedgen discharge` ran qedlift in a tempdir, printed a verdict, and discarded the emitted proof. Now `--out-dir <dir>` **persists** the qedlift artifacts (`<Module>TracedLifted.lean` + `<Module>Refinement.lean`) into the project, so the sorry-free byte-level proof lives where a later `lake build` type-checks it — and where **A2b** can consume it from the bridge.

- `discharge --out-dir` (optional); `run_discharge` gains `out_dir: Option<&Path>`. Omitted ⇒ legacy verdict-only behaviour (back-compat).
- `persist_discharge_artifacts` copies both artifacts into the out-dir (created if needed), overwriting so a re-discharge refreshes in place. Success output reports the persisted paths + the `import <Module>Refinement` wiring line.
- **Scope:** namespace / lakefile-roots auto-wiring stays out of A2a — it's coupled to the bridge consumer (the exact import target + project layout), so it lands with A2b.

## Tests
2 new — the persist helper (copies both, creates a nested dest) and an end-to-end `--out-dir` run through `run_discharge` via a fake qedlift. Full suite green (**1022 bin unit tests + all integration suites**), `cargo fmt --check` + `cargo clippy -- -D warnings` clean.

## Real end-to-end is gated on a qedsvm pin (next unblock)
A *real* discharge needs a `qedlift` that supports `--descriptor`. The descriptor seam is on qedsvm `main` but **post-v0.6.0 and untagged** — `083a3f2` (v1), `880c1ed` (v2 `add_param` — note: actually landed, revising the design note's §16), `d2cc646` (arbitrary literals). v0.6.0's qedlift rejects `--descriptor`. So the next step is a **qedsvm v0.7.0 tag + pin bump**; A2a's persist logic is qedlift-output-agnostic and validated here with a fake qedlift.

### DIFF
--- a/crates/qedgen/src/cli.rs
+++ b/crates/qedgen/src/cli.rs
@@ -396,6 +396,12 @@ pub(crate) enum Commands {
         /// Lean module name for the emitted proof (default: `<Account><Handler>`)
         #[arg(long)]
         module: Option<String>,
+
+        /// Persist the discharged proof (`<Module>TracedLifted.lean` +
+        /// `<Module>Refinement.lean`) into this directory instead of a temp
+        /// dir. Omit to keep the verdict-only (artifact-discarded) behaviour.
+        #[arg(long)]
+        out_dir: Option<PathBuf>,
     },
 
     /// Consolidate multiple proof projects into a single Lean project
--- a/crates/qedgen/src/descriptor.rs
+++ b/crates/qedgen/src/descriptor.rs
@@ -8,7 +8,7 @@
 //! much, never byte offsets. Offsets are *shape*, owned by the IDL and resolved on the qedsvm
 //! side. So qedgen never computes a layout here; it emits pure semantics derived from the spec.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use anyhow::{anyhow, bail, Context, Result};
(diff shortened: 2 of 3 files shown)


abishekk92:
**Real end-to-end validation (local).** Built `qedlift` from qedsvm `main` (`880c1ed`, which has `--descriptor`) and ran the full chain against the committed `vault.so` fixture:

```
qedgen discharge --spec tests/fixtures/descriptor/vault.qedspec \
  --handler increment --account vault \
  --so qedsvm-rs/tests/fixtures/vault.so \
  --idl qedsvm-rs/tests/fixtures/vault.codama.json \
  --qedlift <built qedlift> --out-dir <proj>
→ ✔ DISCHARGED : `increment` is proven against the bytes
```

Both artifacts persisted into `--out-dir`, **0 `sorry`**:
- `VaultIncrementRefinement.lean` — `theorem refines_asm : … AsmRefinesFieldUpdate cr 4 0 0 4 rr baseAddr … := by unfold …` (SPEC-DRIVEN, layout-general) + `theorem ensures : … u64FieldAt 32 [(0,.pubkey owner),(32,.u64 total),(40,.byte bump)] + 1 := …` (offset 32 from `vault.codama.json`).
- `VaultIncrementTracedLifted.lean`.

Confirms the persist plumbing against real qedlift output. CI still gates on the fake-qedlift test since a from-`main` qedlift isn't pinned yet (needs the qedsvm v0.7.0 tag + pin bump).

PR #244, opened 2026-07-17
fix(codegen): wrong-target integration tests, missing proptest dev-dep, account-key effect miscompile

saicharanpogul:
Three reproducible v2.43.0 codegen bugs, all triggered by one minimal Anchor vault spec (`qedgen codegen --all --target anchor` on a spec with a token-account handler and a `field := <signer_account>` effect).

## 1. Anchor target emitted a QuasarSVM integration test

`integration_test::generate` was called for every non-assembly target, but the scaffold is QuasarSVM-shaped end-to-end (`quasar_svm`, `spl_token_interface`, `alloc::vec`, `<name>-client` imports) — an Anchor crate received a `tests/integration_tests.rs` that cannot compile.

- `generate` now takes `Target` and refuses non-Quasar with a clear error.
- The `codegen` arm skips with a note (`--integration`/`--all` on Anchor/Pinocchio) instead of writing a broken artifact.
- regen-drift pins `Target::Quasar` on its file-present path — same reasoning as the existing `Target::Anchor` pin on `kani_impl`.
- `references/cli.md` updated (it claimed `--integration` was target-agnostic).

## 2. Generated Cargo.toml omitted `proptest` from `[dev-dependencies]`

`tests/proptest.rs` imports `proptest::prelude::*` but the generated crate shipped no `[dev-dependencies]`, so the first `cargo test` fails out of the box.

- `render_cargo_toml` now emits `[dev-dependencies]\nproptest = "1"` for all three targets.
- `merge_cargo_toml` dep-merges the section like `[dependencies]`: qedgen upserts `proptest`, user-added dev-deps survive. The owned list is kept separate (`QEDGEN_OWNED_DEV_DEPS`) so a user-added `proptest` in `[dependencies]` is never treated as qedgen-owned.
- Snapshot deltas are exactly the new section (10 fixtures, insertions only).

## 3. `field := <signer>` miscompiled to a bare account name

Assigning a Pubkey state field from a handler account mechanized to `self.vault.admin = new_admin;` — `new_admin` is a free variable in the handler body (E0425), and an account wrapper rather than a `Pubkey` even where it resolves (E0308).

Account-bound RHS (bare `<acct>` and `<acct>.pubkey`) now lowers to the target's runtime key load:

| target | rendered |
|---|---|
| anchor | `self.vault.admin = self.new_admin.key();` |
| quasar | `self.vault.admin = ((*self.new_admin.to_account_view().address())).into();` |
| pinocchio | `*self.new_admin.key()` |

Fixed on both the flat-struct path (`mechanize_effect`) and the destructured multi-variant ADT path (`mechanize_effect_destructured`, disjoint-field borrow so the key read inside the match arm is legal). Arithmetic ops on an account key (`pool += new_admin`) keep falling through to the `todo!()` fill site.

## Validation

- `cargo test` — all 24 test binaries green (7 new unit tests: target gate, dev-dep render + merge semantics, account-key lowering on all targets + both paths)
- `cargo fmt --check`, `cargo clippy -- -D warnings`, `scripts/check-readme-drift.sh` — clean
- End-to-end repro spec re-run on the fixed binary: integration test skipped with note on Anchor / still generated on Quasar; `[dev-dependencies]` present and merge-safe (user deps survive, stale `proptest = "0.9"` upserted); `set_admin` handler compiles as `self.vault.admin = self.new_admin.key();`

Found while spec-driving a vault program on v2.43.0. Also hit (not in this PR): install.sh release-asset skew (installer downloads v2.15.1 assets), partial regen leaving `lib.rs` stale after a spec rename with no `--force`/warning, `(looking in )` empty-path import errors, raw char-class parse errors, and no implication operator in property bodies — happy to file those as issues.


abishekk92 (commented):
Draft feedback — the three reported bugs are valid, and the happy-path fixes work. I reproduced three remaining gaps worth covering before this is marked ready:

1. **Existing wrong-target integration artifacts survive regeneration.** `run.rs` skips generating integration tests for Anchor/Pinocchio, but it does not remove or flag an existing generator-owned Quasar `integration_tests.rs`. `regen_drift.rs` then assumes Quasar whenever that file exists. I reproduced this by running Anchor integration generation against an existing Quasar-shaped file; the file remained unchanged. Please resolve the actual target during regen and either safely remove generator-owned obsolete files or emit a migration error. Add an Anchor upgrade-path regression.

2. **Account-key RHS lowering does not validate the destination type.** A validly parsed spec with `pool : U64` and `effect { pool := new_admin }` emits `self.state.pool = self.new_admin.key();`, which cannot compile. Please gate account-key lowering on a Pubkey-compatible destination, or report a spec diagnostic. Add negative coverage for Anchor, Quasar, and Pinocchio.

3. **Bare account RHS is still unsupported during cross-variant ADT promotion.** With `pragma state_repr = adt`, an `Empty -> Active` transition containing `admin := new_admin` falls back to `todo!()`, although the equivalent `.pubkey` form is recognized by `resolve_cross_variant_rhs`. Please route promotion RHS handling through the typed `ExprTree` / shared `account_key_rhs` resolver and add a cross-variant bare-account regression.

Validation: the new account-key tests passed 7/7, integration target-gate test passed, and Cargo dependency merge tests passed 4/4. The fixes are sound for covered cases; these are completeness gaps.

saicharanpogul:
Thanks for the thorough review — all three completeness gaps are fixed and regression-tested in the latest push.

**1. Obsolete wrong-target integration artifact.** Two-sided fix:
- Codegen (`run.rs`): when targeting Anchor/Pinocchio it now removes a generator-owned `integration_tests.rs` left over from a prior Quasar run (identified by the QEDGEN banner), with a migration note. A user-authored file at that path is preserved with a warning, not deleted.
- `regen_drift`: no longer infers Quasar from file presence — it resolves the real target via `Cargo.toml`/`lib.rs` (`resolve_project_target`, checking both root and nested `programs/`) and only regenerates the integration file for a Quasar project. A leftover in a non-Quasar project now surfaces as drift instead of being silently regenerated.
- Anchor upgrade-path regression added (generate Quasar → regenerate Anchor → file removed), plus `resolve_project_target` and removal-helper unit tests.

**2. Account-key RHS destination type.** `mechanize_effect` / `mechanize_effect_destructured` now gate account-key lowering on a Pubkey destination (`effect_dest_is_pubkey`); a non-Pubkey target (`pool : U64; pool := new_admin`) falls to the `todo!()` path instead of emitting `self.state.pool = self.new_admin.key()`. I also added the spec diagnostic you suggested — a new `effect_type_mismatch` check lint (Error) that names the account, field, and declared type. Negative coverage across Anchor/Quasar/Pinocchio.

**3. Cross-variant bare-account RHS.** `resolve_cross_variant_rhs` now resolves off the typed `ExprTree` via the shared `account_key_rhs`, so the bare `<acct>` and `<acct>.pubkey` spellings unify — `Empty -> Active { admin := new_admin }` lowers to `admin: self.new_admin.key()` instead of bailing to `todo!()`. Same Pubkey-dest gate applies (a bare account into a scalar post field bails). Cross-variant bare-account + bails-on-non-pubkey regressions added.

Full suite green (24 binaries), `fmt`/`clippy -D warnings`/`check-readme-drift`/`check --regen-drift` all clean, snapshots unchanged.

Issue #143, opened 2026-07-02
[qedspec] Kani codegen: ref_impl calls in effect RHS use ML syntax (f (x) (y)) instead of Rust (f(x, y))


xeroc:
# [qedspec] Kani codegen: ref_impl calls in effect RHS use ML syntax (f (x) (y)) instead of Rust (f(x, y))

**qedgen version:** qedgen 2.38.0

**Where the bug surfaces:** `qedgen codegen --kani (Kani harness)`

## Minimal .qedspec

```rust
// Expected: ref_impl calls in effect RHS render as bps_mul(amount, rate).
// Actual:   ref_impl calls render as (bps_mul (amount) (rate)) — ML syntax.
// Note:     property predicates that call the same ref_impl render
//           CORRECTLY with comma syntax. Only the transition-fn effect
//           emitter uses ML application syntax.

spec GenericVault
program_id "11111111111111111111111111111111"

type State = {
  active : U8,
  rate   : U16,
  fee    : U64,
}

type Error | Unauthorized

const BPS_DENOMINATOR = 10000

ref_impl bps_mul (amount : U64) (bps : U16) : U64 =
  mul_div_floor(amount, bps, BPS_DENOMINATOR)

handler execute (amount : U64) : State -> State {
  permissionless
  requires active == 0 else Unauthorized
  effect {
    fee := bps_mul(amount, rate)
  }
}

property fee_correct :
  state.fee == bps_mul(state.fee, state.rate)
  preserved_by all
```

## Commands run

```shell
qedgen init --name repro --spec repro.qedspec
qedgen codegen --spec repro.qedspec --kani --kani-output kani.rs
```

## Expected behavior

```rust
// Transition fn effect RHS:
s.fee = bps_mul(amount, s.rate);

// Property predicate (already correct):
fn fee_correct(s: &State) -> bool {
    s.fee == bps_mul(s.fee, s.rate)
}
```

## Actual behavior

```text
Transition fn:

s.fee = (bps_mul (amount) (rate));    // ← ML application syntax

Property predicate (correct):

fn fee_correct(s: &State) -> bool {
    s.fee == bps_mul(s.fee, s.rate)   // ← correct comma syntax
}

Error: expected `,`, found `(`
```

The ref_impl call renders differently depending on which emitter writes it: property-predicate emitter produces correct Rust; transition-fn effect emitter produces ML syntax.

## Workaround tried

[...]

PR #245, opened 2026-07-17
fix(codegen): read-after-write effect soundness — parallel semantics everywhere — plus guard-fixture solver gaps

saicharanpogul:
Two more findings from dogfooding v2.43.0 on a vault spec, the first one a soundness hole.

## 1. Same-block read-after-write: model and assertion disagreed on pre/post

For `effect { balance += amount, last_seen := balance }`, v2.43's artifacts modeled **two different programs**:

| artifact | `last_seen` gets | semantics |
|---|---|---|
| Lean model (`{ s with balance := s.balance + amount, last_seen := s.balance }`) | pre-add balance | parallel |
| Kani effect-conformance assertion (`s.last_seen == pre_balance`) | pre-add balance | parallel |
| generated handler body | post-add balance | sequential |
| Kani transition fn | post-add balance | sequential |
| proptest transition fn | post-add balance | sequential |
| unit-test apply helper | `s.balance` — unbound, E0425 | (doesn't compile) |

The failure mode: `lake build` proves properties about the parallel model, proptest greens the sequential implementation against its own sequential oracle, and `#[qed(verified)]` stamps the handler — all green while the proof is about a different program. The only artifact that noticed (Kani's conformance harness) failed against **its own transition model**, which reads as a tool bug and gets dismissed.

**Fix**: parallel is canonical — it's what the proofs are already about, and the atomic-transition reading is the spec-language convention (sequential dataflow stays expressible with handler-level `let` bindings). Every sequential emitter now snapshots fields the block both writes and reads (`let pre_<field> = …;`) and routes RHS reads through the snapshot:

- kani + proptest transition fns (one shared emitter, `emit_transition_fn_inner`)
- generated handler bodies — flat `mechanize_effect` and the destructured multi-variant ADT path (the cross-variant promotion path was already parallel by construction)
- unit-test apply helpers and effect assertions

Snapshots are emitted only for fields actually referenced, so specs without read-after-write generate byte-identical output — the full snapshot suite and the bundled examples pass unchanged, which is also why this survived: the bundled corpus never exercises the shape.

`references/qedspec-dsl.md` now states the semantics explicitly (new "Effect semantics: parallel (pre-state) reads" section) with the `let`-binding pattern for sequential dataflow. The #143–146 regression fixture's expectations updated to the parallel form (its `cut := bps_mul(fee, rate)` read-after-write now renders `pre_fee`, matching what its own Lean model always meant).

## 2. Unit-test generator: `s.` leak + guard fixtures that violate their own guard

- **`s.<field>` E0425**: apply helpers and effect assertions rendered state-field RHS with the harness-model `s.` binder inside `state.`-receiver scopes — `state.last_seen = s.balance;` doesn't compile. Now rendered against the `state` binder.
- **Bool requires ignored**: bool fields always seeded `false`, so `requires seat_open == true` produced a `test_X_guard_accepts_valid` that fails on correct code. Bool atoms (`f == true/false`, bare `f`, `!f`) now pin the fixture field, and the rejects-test can violate a bool-only guard by flipping it.
- **Cross-field sum clauses dropped**: `requires amount + last_seen <= cap` resolved to `None` and was skipped, leaving `cap: 0` in the fixture. `+`-sums now participate in the raise fixpoint (value-bearing, never the adjustment target).

## Validation

- End-to-end repro spec: generated unit tests now compile and pass 10/10 against the generated apply helpers (v2.43: file didn't compile); the kani transition satisfies its own conformance assertion; handler bodies match the Lean model; both guard fixtures satisfy their guards.
- `cargo test` — all 24 test binaries green, including 7 new regression tests (transition snapshot + no-op cases, substitution token boundaries, mechanize substitution, apply/assert receiver, bool + sum fixtures, bool-only rejects).
- `cargo fmt --check`, `cargo clippy -- -D warnings`, `scripts/check-readme-drift.sh`, `check --regen-drift` (bundled examples byte-identical) — all clean.

Branched from upstream/main, independent of #244 — both touch `codegen_shared/effect.rs`, so whichever merges second needs a trivial rebase.

Known edge left open (happy to file as an issue): duplicate writes to the same field in one block (`effect { b += 1, b += 2 }`) are ill-defined under parallel semantics and currently last-write-wins in Lean vs. cumulative in Rust — a lint rejecting duplicate targets would close it structurally.


abishekk92 (commented):
Draft feedback — the reported artifact-semantics bug is real, and the pre-state snapshot fix works for the normal one-write-per-field case. I reproduced two valid DSL shapes that still break generated artifacts:

1. **Duplicate effect targets still diverge from the documented parallel semantics.** For `effect { b += 1; b += 2 }`, Lean emits parallel record updates based on pre-state, while generated Rust applies the operations sequentially and produces `pre_b + 3`. The generated unit test is internally contradictory and fails. Please either reject duplicate normalized LHS targets during checking or lower them through pre-state temporaries with one final commit per target. Add an end-to-end duplicate-target regression.

2. **Rejecting fixtures are incorrect for disjunctive boolean guards.** For `requires enabled == true or emergency == true`, generation seeds both fields `true`, flips only one atom (or in my reproduction leaves both true), and then asserts the OR guard is false. The generated test fails against correct guard logic. `split_atoms` currently discards the `&&`/`||` structure. Please preserve and negate the complete guard AST; an OR is rejected only when every disjunct is false. Add boolean-OR and nested AND/OR regressions.

Validation: targeted parallel-snapshot tests passed 3/3 and the existing bool/sum fixture test passed. A standalone generated regression suite passed 6/8; exactly the duplicate-target effect test and OR-guard rejection test failed.

saicharanpogul:
Thanks — both remaining shapes are fixed and regression-tested in the latest push.

**1. Duplicate effect targets.** I took the "reject during checking" option (it matches the "eliminate the proof obligation structurally" doctrine, and lowering-with-one-commit would still have to pick a semantics that silently disagrees with what the author likely meant). `check_effect_targets` now bails on a block that writes the same normalized target twice — across all backends — and a new `duplicate_effect_target` completeness lint (Error) surfaces it at check time with a fix suggestion (`b += <total>`). Crucially it's checked **per effect block**: `match` arms are independent, so mutually-exclusive writes to one field don't false-positive. This also closes #256, which I'd filed as a known edge — thanks for pushing to fix it in-PR rather than defer.

**2. OR-guard rejection fixtures.** You were exactly right that `split_atoms` threw away the boolean structure. Replaced the single-atom negation with a structure-aware solver over the typed `requires` trees (`solve_tree`): falsifying an AND needs one conjunct false, an OR needs **every** disjunct false, with mutual satisfy/falsify recursion through `not` and arbitrary nesting. For `requires enabled == true or emergency == true` the rejects fixture now sets both `false`; for `(a > 0 or b > 0) and c > 0` it falsifies a full conjunct. The generated OR/nested tests compile and pass against the real predicate (verified end-to-end, not just by string match). Bool-OR, numeric-OR, and nested-AND/OR regressions added; the legacy single-atom path stays as a fallback.

Full suite green (24 binaries), `fmt`/`clippy -D warnings`/`check-readme-drift`/`check --regen-drift` clean, snapshots unchanged.

PR #259, opened 2026-07-18
feat(adapt): fill accounts{} + State from IDL-derivable info in scaffold skeleton (#257, #258)

### LINKED ISSUE #257: adapt/scaffold skeleton leaves accounts{} + signer as TODO though the IDL declares every account (signer/writable/pda)
## Summary

The scaffold-to-spec skeleton (`qedgen adapt` / `qedgen probe --program --emit-spec-candidates`) renders **every** handler with a placeholder accounts block and a placeholder signer:

```
handler create_transaction (args : CreateTransactionArgs) : State.Init -> State.Init {
  // TODO: auth <signer>
  // TODO: accounts { ... }
  // TODO: requires
  // TODO: effect { ... }
}
```

But for an Anchor program the source IDL (and the `#[derive(Accounts)]` struct) declares, per instruction, **every account with its `isMut` / `isSigner` / `pda` flags** — and the accounts-struct name is already captured. This is mechanically derivable and should be emitted, not left as a `TODO`. Guiding principle: anything mechanically derivable from the IDL should not be a `TODO` in the generated skeleton.

## Evidence

- `crates/qedgen/src/adapt/anchor_adapt/render.rs:106-111` captures only the Accounts struct **name** and renders it as a `/// accounts struct: ...` doc comment.
- `render.rs:139-140` then emits static `// TODO: auth <signer>` and `// TODO: accounts { ... }`.
- The handler `entry` carries `accounts_type` (name only) but no per-account flags — `anchor_extractor.rs` does not thread the account list.
- A representative IDL instruction carries the full list, e.g. `create_transaction`: `settings` (writable), `transaction` (writable), `creator` (signer), `rent_payer` (writable, signer), `system_program` (read-only). `approve_proposal`: `settings` (ro), `signer` (writable, signer), `proposal` (writable), `system_program` (ro).

## Proposed fix

1. `anchor_extractor.rs`: capture each instruction's account list (name + `isMut`/`isSigner`/`pda`) from the IDL (fall back to the `#[derive(Accounts)]` struct when no IDL).
2. `render.rs`: emit the real `accounts { <name> : signer, writable, ... }` block instead of `// TODO: accounts { ... }`, and seed `auth` from the declared signer name(s) instead of `// TODO: auth <signer>`.

Skeleton stays a skeleton — `requires` / `effect` still need the impl and remain `TODO` — but the mechanical account/signer facts are filled. Additive; add a snapshot test over a small fixture with a multi-account instruction. Surfaced dogfooding an audit of a large brownfield Anchor program (39 handlers): 21/37 handlers were left as bare `Init -> State.Init` shells with `TODO` accounts despite the IDL declaring all of it.

### LINKED ISSUE #258: adapt/scaffold skeleton emits flat State.Init->State.Init though account structs carry status enums (seed multi-variant State from IDL)
## Summary

The scaffold-to-spec skeleton hardcodes `: State.Init -> State.Init` for every handler and emits a flat placeholder `type State | Init | Active`, even when the IDL's account structs carry **status enums** that are the program's real lifecycle state machine.

For example, an Anchor multisig's IDL declares `Proposal.status : ProposalStatus` with variants `Draft | Active | Rejected | Approved | Executing | Executed | Cancelled`. The state-machine **variants** are mechanically derivable from the IDL; only the per-handler transition **edges** (which `from -> to` each handler takes) require the impl. Today the skeleton throws all of it away and emits a flat self-loop.

## Evidence

- `crates/qedgen/src/adapt/anchor_adapt/render.rs:128` hardcodes `s.push_str(" : State.Init -> State.Init {\n");` for every handler.
- The generated `type State` is a two-variant `Init | Active` placeholder unrelated to the program.
- The IDL carries the enums: `types[]` includes `ProposalStatus` (7 variants), `SettingsAction`, `Permission (Initiate|Vote|Execute)`, `Period`; and `accounts[]` includes `Proposal.status : ProposalStatus` (an account field typed by the enum).

## Proposed fix

1. `anchor_extractor.rs`: read account status-enum fields and their variants from the IDL (`accounts[].type.fields[]` typed `defined: <Enum>`, resolved against `types[]`).
2. Skeleton emits a real `type State | Draft | Active | Rejected | Approved | Executing | Executed | Cancelled` (per account with a status field), and renders each handler's transition as `: State.? -> State.?  // one of: Draft|Active|...` so the author fills the *edge* against real variants instead of inventing the state type.

This seeds the state **type** mechanically; the transition edge and effect still need the impl (not IDL-derivable) and stay `TODO`. Additive; snapshot test over a fixture whose account carries a status enum. Same dogfooding origin as the sibling accounts-block issue — the flat `Init -> Init` shell was the reason a large-lifecycle Anchor program's handlers came out unverifiable from the skeleton alone.

abishekk92:
Closes #257. Closes #258.

Two mechanically-derivable facts the scaffold-to-spec skeleton was dropping to `TODO`/placeholder, from the guiding principle: **anything mechanically derivable from the IDL / `#[derive(Accounts)]` / account structs should not be a `TODO`.**

## #257 — fill `accounts { }` from the handler's `#[derive(Accounts)]` struct

Was `// TODO: accounts { ... }`. Now each handler renders a real block: `Signer` → `signer`, `#[account(mut)]` → `writable`, `Program`/`Sysvar` → `program`, `Account<T>`/`InterfaceAccount<T>`/`Box<Account<T>>` (read-only) → `type T`. The declared signer is surfaced as an `auth` hint comment (a live `auth <signer>` trips the `unbound_auth` P1 lint on a fresh skeleton; the signer fact is already in the block).

## #258 — seed `type State` from an `#[account]` status enum

Was a flat `type State | Init | Active` placeholder. Now, when an `#[account]` struct carries a program-defined status enum (e.g. `Proposal.status: ProposalStatus`), the skeleton emits the real lifecycle (`Init` retained as the pre-existence placeholder so `State.Init` transitions stay valid): `Init | Draft | Active | Rejected | Approved | Executing | Executed | Cancelled`. Two passes over `src/`, preferring a `status`/`state` field and the richest enum. Only the variant set is derived — transition edges need the impl, so the variants are unreachable until wired and `check` emits the expected P2 `lifecycle_unreachable_state` hints.

## Impact

On a large brownfield Anchor program (39 handlers): 0 → 39/39 handlers with real `accounts {}` blocks, and the flat State placeholder becomes the actual 7-state proposal lifecycle.

## Scope / follow-ups

- PDA seed rendering deferred (grammar needs `pda [seeds]`; translating Anchor `seeds = [...]` is error-prone) — a mut PDA renders as `writable` for now.
- Only the primary status enum seeds State; multi-state-machine programs get the richest one.

## Changes

`program_model` (`AccountRoleModel`, `StateModel`, `HandlerModel.accounts`, `ProgramModel.state`); `anchor_adapt/{extract,render,adapter,mod}`; 3 fixtures regenerated; 4 new tests.

## Verification

Full suite: 1406 passed, 0 failed. clippy `--all-targets`: clean. rustfmt: clean.


abishekk92 (commented):
The account/state derivation direction is good, but these issues should be fixed before merge:

1. **Qualified Accounts paths resolve to the wrong struct.** `resolve_account_roles` discards the qualified `Context<crate::b::Shared>` path and selects the first bare-name match across `src/`. The existing `anchor-accounts-collision` fixture reproduces this: both handlers render the first `Shared` layout, while the second handler should use `b::Shared`. Please reuse the qualified-path resolver already used in `attributes.rs` and add a scaffold regression.

2. **`#[account(signer)]` is not recognized.** Signer classification only checks `Signer<'info>`, so valid signer constraints on `AccountInfo` or `UncheckedAccount` are omitted or rendered merely writable. Please combine type-based detection with `account_attr_has_word(..., "signer")` and test both `signer` and `mut, signer`.

3. **Required CI is red.** Stable Clippy rejects the `match inner_ty` block in `extract.rs` under `-D warnings` (`clippy::question_mark`), so the test and downstream gates never ran. Replace it with `let ty = inner_ty?;` or otherwise address the lint.

Validation: targeted Anchor adapter tests passed 27/27, and the qualified-path collision was reproduced. Supply-chain CI passes; the required test job fails at Clippy.

PR #263, opened 2026-07-18
fix(codegen): five v2.45.0 codegen-quality gaps (bool fixtures, chained comparison, PDA signer, compound guard solver, init/space)

saicharanpogul:
Five codegen-quality gaps found dogfooding a Fortel-vault-shaped spec against v2.45.0 (a state machine with a `Bool` field, an `Uninitialized → Active` PDA init, a bool-from-comparison effect, a vault-authority token CPI, and compound cross-field requires). None are soundness holes — all are generated artifacts that don't compile or don't run — each with an exact emitter root cause and a regression test.

## 1. Bool fixtures emit `0` in proptest/kani (`codegen_shared/types.rs`)
`default_value_for_type` fell through to the numeric `"0"` catch-all for `Bool`, so the proptest/kani seeders emitted `seat_active: 0` → `E0308 expected bool`. The unit-test seeder special-cased it locally. Fixed at the shared source (`Bool → "false"`).

## 2. Chained comparison in the Kani conformance assert (`kani_mir/conformance.rs`)
A `set` effect whose RHS is a comparison (`seat_active := seat_stake > 0`) rendered `assert!(s.seat_active == pre_seat_stake > 0, …)` — `error: comparison operators cannot be chained`. The assert is built by string concat, not `render_cmp`, so it's parenthesized here when the RHS tree is a `Cmp`/`BoolOp`.

## 3. PDA-signer seeds missing on vault-authority CPI (`codegen_shared/cpi.rs`)
`call Token.transfer(authority = vault /*PDA*/, …)` emitted `CpiContext::new(...)`, which fails at runtime with a missing-signer/privilege error — a PDA authority needs `invoke_signed`. When the call's `authority` account is a program PDA, codegen now assembles its `#[account(seeds = …, bump)]` seeds and emits `CpiContext::new_with_signer(…, signer_seeds)`, binding the seed account keys to locals first (`.key()` returns an owned `Pubkey` temporary — E0716 otherwise). A non-PDA authority keeps plain `new`. The seeds mirror the account's own PDA constraint, so they're deterministic.

## 4. Guard-fixture solver dropped compound cross-field constraints (`codegen/unit_test.rs`)
`stake_slash + lp_loss == loss` (linear equality over params), `amount > collateral` (field-vs-field), and disjunctions (`lp_loss == 0 or …`) were all dropped, so the accepts/rejects fixtures violated their own guard and the generated `assert!` failed against correct code. Added a bounded constraint-propagation solver over the typed `requires` trees:
- **satisfy** (accepts): linear-equality solving + inequality assignment + disjunction enforcement, VERIFIED before use.
- **falsify** (rejects): make one clause false, handling field-vs-field and linear forms, VERIFIED.
- **param-shadow aware**: a param that shadows a same-named state field (the guard fn reads the param) is no longer conflated with the field's seed.
- **safety net**: an un-auto-solvable guard smoke-tests the guard fn (`let _ = guard_…`) instead of emitting a failing assertion — a generated test never fails on correct code.

All bundled-example unit tests now compile and pass (previously lending 1 + multisig 5 failing, latent because CI doesn't compile+run them). 19 of 24 example guard tests are now *solved* with real assertions; only multisig's 5 Map-threshold guards fall back to honest smoke-tests.

## 5. init/payer/space not emitted on `Uninitialized → Active` (`codegen_shared/account_attr.rs`)
Reported as an investigation ("logic present, not emitted"). Root cause: a type-qualified pre-state (`State.Uninitialized`) sets `on_account = Some("State")` — the state TYPE — and the `on_account_matches` name heuristic then rejected any PDA not named `state` (e.g. `vault`), so `init/payer/space` never emitted. Three-part fix:
- Gate init on the resolved `is_state_account` for single-account specs (the type qualifier isn't a per-account discriminator there).
- Fix the `space` struct name: it built `StateAccount::INIT_SPACE` from the type qualifier, an undefined type — now the real `<Program>Account`.
- Derive `InitSpace` on the flat Anchor `#[account]` struct (the multi-variant ADT wrapper already did) so the emitted init scaffold compiles — this was latent because init was never emitted before.

## Validation
- `cargo test` — full suite green, incl. 8 new regression tests (bool default, conformance parens, PDA signer + non-PDA, compound accepts/rejects, param-shadow, unsolvable→smoke-test, init emission + non-init).
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `scripts/check-readme-drift.sh`, `check --regen-drift`, and `codegen_smoke --ignored` (Anchor scaffolds compile) all clean.
- Bundled example unit tests compile and pass end-to-end.

Branched off v2.45.0 (`upstream/main`, after #244/#245 merged).

### DIFF
--- a/crates/qedgen/src/codegen/codegen_mir.rs
+++ b/crates/qedgen/src/codegen/codegen_mir.rs
@@ -2058,6 +2058,112 @@ handler poke : State.Active -> State.Active {
         );
     }
 
+    /// v2.46 (Bug 5) — a single-ADT `State.Uninitialized -> State.Active`
+    /// handler over a PDA must emit `#[account(init, payer, space = 8 +
+    /// <Program>Account::INIT_SPACE, …)]`. The type-qualified pre-state
+    /// (`State.`) previously set `on_account = Some("State")`, and the
+    /// name heuristic then rejected a PDA named anything but `state`, so
+    /// init/payer/space never emitted (and the space struct name resolved
+    /// to the non-existent `StateAccount`).
+    #[test]
+    fn anchor_init_emits_for_type_qualified_single_adt_pda() {
+        let src = r#"spec Vault
+program_id "11111111111111111111111111111111"
+type State
+  | Uninitialized
+  | Active of { admin : Pubkey, balance : U64 }
+type Error | Bad
+pda vault ["vault", admin]
+handler initialize : State.Uninitialized -> State.Active {
+  accounts {
+    admin : signer, writable
+    vault : writable, pda ["vault", admin]
+    system_program : program
+  }
+  effect { state := .Active { admin := admin, balance := 0 } }
+}
+"#;
+        let parsed = crate::chumsky_adapter::parse_str(src).expect("parse");
+        let mir = crate::mir::lower(&parsed);
+        let fp = crate::fingerprint::compute_fingerprint(&parsed);
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let temp = tmp.path();
+        std::fs::create_dir_all(temp.join("src")).expect("mk src");
+        emit_lib(
+            &mir,
+            &parsed,
+            &fp,
+            temp,
+            Target::Anchor,
+            Path::new("unused.qedspec"),
+            RegenOptions::default(),
+        )
+        .expect("emit lib");
+        let lib = std::fs::read_to_string(temp.join("src/lib.rs")).expect("lib.rs");
+        // The vault line must carry init/payer/space with the correct
+        // wrapper struct name (`VaultAccount`, not `StateAccount`).
+        let vault_line = lib
+            .lines()
+            .find(|l| l.contains("seeds = [b\"vault\""))
+            .unwrap_or_else(|| panic!("no vault #[account] line in:\n{lib}"));
+        assert!(
+            vault_line.contains("init")
+                && vault_line.contains("payer = admin")
+                && vault_line.contains("space = 8 + VaultAccount::INIT_SPACE"),
+            "init/payer/space must emit with the VaultAccount wrapper; got:\n{vault_line}"
+        );
+        assert!(
+            !lib.contains("StateAccount::INIT_SPACE"),
+            "space must not reference the non-existent StateAccount; got:\n{lib}"
+        );
+    }
+
+    /// A non-init handler on the same PDA stays `mut` (init only on the
+    /// Uninitialized→X transition).
+    #[test]
+    fn anchor_non_init_handler_keeps_mut_not_init() {
+        let src = r#"spec Vault
+program_id "11111111111111111111111111111111"
+type State
+  | Uninitialized
+  | Active of { admin : Pubkey, balance : U64 }
+type Error | Bad | MathOverflow
+pda vault ["vault", admin]
+handler deposit (amount : U64) : State.Active -> State.Active {
+  accounts {
+    admin : signer
+    vault : writable, pda ["vault", admin]
+  }
+  effect { balance += amount }
+}
+"#;
+        let parsed = crate::chumsky_adapter::parse_str(src).expect("parse");
+        let mir = crate::mir::lower(&parsed);
+        let fp = crate::fingerprint::compute_fingerprint(&parsed);
+        let tmp = tempfile::tempdir().expect("tempdir");
+        let temp = tmp.path();
+        std::fs::create_dir_all(temp.join("src")).expect("mk src");
+        emit_lib(
+            &mir,
+            &parsed,
+            &fp,
+            temp,
+            Target::Anchor,
+            Path::new("unused.qedspec"),
+            RegenOptions::default(),
+        )
+        .expect("emit lib");
+        let lib = std::fs::read_to_string(temp.join("src/lib.rs")).expect("lib.rs");
+        let vault_line = lib
+            .lines()
+            .find(|l| l.contains("seeds = [b\"vault\""))
+            .expect("vault line");
+        assert!(
+            vault_line.contains("mut") && !vault_line.contains("init"),
+            "non-init handler must keep mut, not init; got:\n{vault_line}"
+        );
+    }
+
     /// The disambiguation is scoped: a spec with no prelude-colliding type
     /// gets no explicit re-import line.
     #[test]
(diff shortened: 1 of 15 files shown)


abishekk92 (changes requested):
Thanks for fixing these codegen problems. The changes are useful, but please fix these three issues before merge.

1. **High: the PR does not compile with current `main`.** Two new tests call `emit_lib` with six arguments. It now needs a seventh `RegenOptions` argument. Add `RegenOptions::default()` to both calls.

2. **High: an unverified guard test can still fail.** The fallback test still runs the guard with an unverified fixture. A division or modulo guard can get a zero value and panic even when the generated program is correct. Do not run the guard with an unverified fixture. Also, make `eval_num_tree` return `None` for division or modulo by zero. Add a regression test.

3. **Medium: the solver can overflow on `U128` values.** It converts `u128` to `i128` with `as` and then uses unchecked arithmetic. Values above `i128::MAX` can wrap or panic during code generation. Use checked conversion and checked arithmetic. Fall back when the value is outside the solver range. Add a large-`U128` regression test.

I tested the PR after adding only the missing `RegenOptions::default()` arguments. All 1,385 unit tests passed. The full workspace tests, formatting check, and Clippy also passed.

abishekk92:
The #294 P0 executable generated-artifact gate independently reproduced the class-4 guard-fixture failures from clean regenerations of the bundled examples (this tree does not yet carry the #263 fixes):

- multisig `tests/unit.rs`: 5/51 FAIL — `test_add_member_guard_rejects_invalid`, `test_approve_guard_rejects_invalid`, `test_reject_guard_rejects_invalid`, `test_execute_guard_rejects_invalid`, `test_cancel_proposal_guard_accepts_valid` (46 pass).
- lending `tests/unit.rs`: 1/22 FAIL — `test_liquidate_guard_rejects_invalid` (21 pass).

Counts match this issue's 'previously lending 1 + multisig 5 failing' claim exactly. When the class-4 fix lands, the gate serves as the executable acceptance check. Cross-referenced in docs/toolchain-backlog.md (session 2026-07-20).

saicharanpogul:
Thanks for the careful review — all three are addressed. Rebased onto current `main` (now past #315 and #316), pushed as `d90a663`.

**1. Compile against `main` (RegenOptions).** Done. Both Bug-5 tests now pass `RegenOptions::default()` as the 7th `emit_lib` arg. `cargo build` / `cargo test` are clean on the rebased base.

**2. Never run the guard with an unverified fixture.** Fixed properly, not just papered over. Two parts:
- `eval_num_tree` now returns `None` on division/modulo by zero (`checked_div`/`checked_rem`), so a trial assignment that would divide by zero is never treated as verified.
- Guard-test emission is restructured: `check_fixture` is computed *before* anything is emitted, and only a `Solved` fixture emits the fixture + `assert!`. An unverified fixture (`Contradicted`/`Unsupported`) now emits **only** `let _ = guard_X;` — a reference that type-checks the guard fn but never calls it. Previously (via #315's `emit_guard_assertion`) the fallback emitted `let _ = guard_X(&state…)`, which *executes* the guard — exactly the div/mod-by-zero panic you flagged. That path is gone; `grep 'let _ = guard_[a-z_]*(&'` across the regenerated examples returns nothing.
- Regression test `division_guard_smoke_tests_without_executing_the_guard` asserts the accept side of a `amount / rate` guard (rate defaults to 0) takes the non-executing path and never calls the guard; `eval_num_tree_returns_none_on_div_by_zero` covers the eval change directly.

**3. U128 overflow in the solver.** Fixed. `u128 → i128` now goes through `i128::try_from(...).ok()?` (falls back to unsolved when out of range), and every arithmetic step is checked (`checked_add`/`checked_mul`/`checked_neg`/…), returning `None` on overflow so the guard smoke-tests instead of wrapping or panicking. Override extraction uses `u128::try_from(...).ok()`. Regression test `large_u128_guard_does_not_panic_and_smoke_tests` drives a `2^127 + 1` literal (above `i128::MAX`) through codegen and confirms no panic; `eval_num_tree_checked_arithmetic_overflows_to_none` covers the arithmetic path.

**On the overlap with #315/#316:** #315 landed the verify-before-assert framework independently while this was in review, and #316 refactored account emission into `AccountPlan`. I kept both and composed on top — my linear solver is now the *fixture source* feeding #315's `check_fixture`, so compound cross-field guards (`a + b == total`, field-vs-field, disjunctions) get a real satisfying/violating fixture and **assert** rather than smoke-test, and Bug 5's init/space gate fix moved into `AccountPlan::derive` (the type-qualified `State.Uninitialized` case #316 still resolved by name only). Verified end-to-end: a single-account vault spec now emits `#[account(init, payer = creator, space = 8 + VaultAccount::INIT_SPACE, …)]`.

Local gates all green: `cargo test`, `fmt --check`, `clippy --all-targets -D warnings`, `check --regen-drift`, `check-readme-drift.sh`, `release-gate.sh`, and `codegen_smoke --ignored`.

abishekk92 (changes requested):
Two of the three requested fixes are resolved: the PR compiles on current main, and unverified guard fixtures now reference the guard function without executing it. The checked arithmetic in the new solver is also good, but the large-U128 path is still unsound in the final fixture verifier.

`eval_tree` still has `ExprTree::Int(v) => Some(FixtureValue::Int(*v as i128))`. For the added `2^127 + 1` regression, that wraps the literal negative. `check_fixture` then incorrectly accepts the fixture and generates:

```rust
let amount: u128 = 100;
assert!(guard_bump(&state, amount));
```

That assertion is false under the actual u128 guard. I reproduced this by strengthening `large_u128_guard_does_not_panic_and_smoke_tests` to require `let _ = guard_bump;`; it fails and prints the assertion above. The submitted regression currently allows either smoke-testing or asserting, so it misses the incorrect assertion.

Please change the `eval_tree` literal conversion to `i128::try_from(*v).ok().map(FixtureValue::Int)` (and audit any equivalent verifier conversion), then make the regression require the non-asserting smoke-test path for the out-of-range literal.

abishekk92 (approved):
All requested changes are resolved. The out-of-range U128 fixture path now uses checked conversion, the regression requires the safe non-executing smoke-test path, local focused tests pass, and all four CI checks are green.

Issue #332, opened 2026-07-22
verify: backend-obligation manifest — every requested obligation ends emitted, unsupported(reason), or failed(reason)


abishekk92:
## Problem

Obligations can vanish silently between the spec and a backend. Four filed issues are instances of this one defect class:

- #324 — Kani multi-account specs drop `cover`/`liveness`/`environment` obligations with no trace.
- #326 — Kani verifies a flat state model for `state_repr = adt` specs and nothing reports the representation gap.
- #328 — the Lean backend drops account-pubkey authorization clauses from guards and abort theorems.
- #331 — multi-account proptest omits ghost-referencing properties behind a non-gating warning.

Each of those issues names the same interim safeguard: report the skip as a structured, machine-readable capability result. This issue owns the design of that mechanism so the four issues do not each half-specify it and drift.

The deep fixes (product-state lowering, Lean context threading, ADT Kani parity) are tracked in the child issues. This issue is only the manifest and its reporting surfaces.

## Coverage taxonomy

The manifest must let every surface distinguish three levels (from #324):

1. **spec coverage** — a handler is named by a property or file-level obligation;
2. **backend coverage** — the selected backend emitted a faithful obligation;
3. **execution coverage** — the generated artifact compiled and the verifier ran it.

The current coverage matrix reports only the first level. The gaps above all live between level 1 and level 2.

## Contract

Codegen produces a backend-obligation manifest from the lowered spec. Every requested obligation for a selected backend must finish as exactly one of:

1. `emitted` — with the generated harness/theorem/test identifier;
2. `unsupported(reason)` — with a machine-readable capability reason (for example `kani_multi_account_file_level`, `lean_handler_account_pubkey`, `proptest_multi_account_ghost`, `kani_adt_state_repr`);
3. `failed(reason)` — lowering or generation should have succeeded and did not.

Rules:

- The manifest is derived from the lowered spec, not from scanning generated files. Generated comments (`// liveness <name>: skipped — …`) remain useful evidence but are not the source of truth.
- The status type is a closed enum. All accounting goes through exhaustive `match` (the `SeverityCounts::of` discipline), so a new status or reason variant is a compile error at every consumer, not a silently dropped case (#260/#270).
- An obligation key identifies (backend, obligation kind, handler or file-level scope, source span). The same spec obligation appears once per selected backend.

## Surfaces

- `qedgen codegen` writes the manifest (suggested location: `.qed/obligations.json`, one per generation run) and prints a summary line per backend.
[...]

Issue #326, opened 2026-07-22
kani: state_repr = adt still verifies a flat state model


abishekk92:
## What happens

`state_repr_is_adt()` is consumed by Lean codegen and scaffold helpers, but not by `kani_mir`. A spec with `pragma state_repr = adt` therefore gets an inductive Lean state model and a flat union-of-fields Kani `struct State` model.

## Why it matters

This is not only a presentation difference. The flat Kani state domain can contain combinations of fields that no ADT variant can inhabit. Proving over that larger domain may produce false counterexamples, and transition semantics are not automatically equivalent to the inductive Lean model.

The mismatch does not by itself make a successful universal Kani proof unsound, but users must not read Kani and Lean results as representation-parity evidence when they are proving over different state spaces.

## Long-term fix

Teach the Kani MIR backend to consume the resolved state representation and emit a discriminated Rust enum (or an equivalent tagged product with an enforced validity invariant) for `state_repr = adt`. Handler transitions must pattern-match only their legal source variants and construct the declared target variants.

Until representation parity lands, report the mismatch through the #332 backend-obligation manifest (reason `kani_adt_state_repr`), surfaced in `check --coverage`, generated artifacts, and `verify --kani`; documentation is necessary but not sufficient.

## Acceptance criteria

- `state_repr = adt` is visible to Kani lowering.
- Kani cannot generate impossible cross-variant field combinations as valid states.
- Transition harnesses preserve declared source/target variant semantics.
- The backend report states the actual representation used.
- ADT fixtures compare Lean and Kani obligation inventories and exercise variant-mismatch paths.
- `docs/framework-support.md` and the DSL pragma reference explain the temporary compatibility behavior until parity ships.

PR #343, opened 2026-07-22
feat(lean): ActionCtx binds account reads — authorization clauses stop dropping (#328)

### LINKED ISSUE #328: lean: model handler-account pubkeys instead of dropping authorization clauses
## What happens

Any `requires`/abort predicate that mentions a handler account's pubkey is skipped in two places in the Lean backend:

- the transition guard builder drops it from the `if` condition via `mentions_handler_account_pubkey(...) { continue; }`;
- the abort-theorem emitter drops the corresponding theorem with the same predicate check.

The generated `Spec.lean` carries no obligation marker and no backend-coverage status. Account-pubkey authorization therefore remains a Rust runtime check while the Lean transition relation silently becomes weaker.

## Why it is this way

The referenced account pubkey is an input to the handler invocation, not a `State` field, and the current Lean transition signature does not bind it. Avoiding a free Lean identifier is correct; deleting the semantic guard is not an acceptable final model.

## Long-term fix

Represent handler invocation context explicitly in Lean. Generate a typed per-handler input/context structure—or explicit parameters—containing every signer, account pubkey, and other non-state value referenced by guards, effects, and abort predicates.

Transitions and their theorems should consume the same context:

```lean
def applyAction (s : State) (ctx : ActionCtx) (amount : Nat) : Option State := ...
```

Then lower account-pubkey predicates normally and emit abort theorems over `ctx`. This scales beyond adding one ad hoc `Pubkey` parameter whenever another free invocation value appears.

Until context threading lands, the #332 backend-obligation manifest must mark each dropped authorization clause `unsupported(reason)` (reason `lean_handler_account_pubkey`), generated Lean should contain a structured marker, and strict verification must not report a fully passed Lean backend. Those transparency measures are interim safeguards, not closure for this issue.

## Acceptance criteria

- Every account/key reference used by a handler predicate is bound in the generated Lean transition signature.
- The transition guard contains the authorization predicate rather than dropping it.
- A corresponding abort theorem is emitted and elaborates.
- Dotted `auth`, explicit `.pubkey`, `.key()`, zero/one/multiple signer handlers, and imported-account field authorization have fixtures.
- `lake build` passes for all supported fixtures.
- Obligation inventory tests prove that every parsed `requires ... else ...` is emitted or explicitly unsupported—never absent.

abishekk92:
Independent of the #326/#337 Kani stack — cut from main. Whichever of this and #338/#339 merges second may need a trivial `spec_tests.rs` / `obligations/mod.rs` conflict resolution.

## Problem

Any `requires`/abort predicate reading a handler account (`.pubkey`, `.key()`, or an imported state field like `admin_config.admin`) was silently dropped from the Lean transition guard, and its abort theorem skipped. Account-pubkey authorization stayed a Rust runtime check while the Lean transition relation quietly weakened. The #332 manifest made this visible (`unsupported(lean_handler_account_pubkey)`); this PR closes the gap for flat shapes.

## Fix — explicit invocation context

- **`structure ActionCtx`** — one spec-level structure carrying every account address and imported cross-program state field the guards read (union across handlers, typed through `mir.imports`). Emitted only when some guard reads one; specs without such reads are byte-identical.
- **Transitions that read any take `(ctx : ActionCtx)`** and keep the clause in the guard: `ctx.initializer_ta = s.initializer_token_account`. Rendering is tree-structural (`LeanCx::with_action_ctx`): `<acct>`/`<acct>.pubkey` → `ctx.<acct>`, `<acct>.<field>` → `ctx.<acct>_<field>`.
- **One shared decision point** (`render_guard_pred`) drives both the transition guard builder and the abort emitter, so the guard conjunct and theorem hypothesis stay byte-identical and the mechanical `if_neg` abort proofs still close — escrow's two previously-dropped `Unauthorized` aborts now **prove, no sorry**.
- **ctx threads through every consumer**: `applyOp`/`applyOps`, property sub-lemmas and master inductive theorems (delegation passes ctx), ensures, frames, overflow, covers (existential ctx witness; closed-term witness traces fall back to `sorry`), liveness (universally bound; ctx-reading paths keep the `sorry` form), and multi-account per-account blocks (`PoolActionCtx` via the rename table).
- **Honest scope**: flat-state shapes with single-projection reads. ADT and indexed lanes, deeper projections, and tree-less carriers keep drop-and-report.

## Validation

- Regenerated escrow `Spec.lean` builds clean under `lake build` — Spec **and the user-owned Proofs.lean** (303 jobs). The only `sorry` warnings are the pre-existing cover/liveness fallbacks.
- Escrow flips to **10 emitted / 0 unsupported** on the Lean backend.
- Snapshot delta is escrow-only; ADT / indexed / multi-account fixtures byte-identical.
- Full suite green (35/35); clippy and fmt clean.

## Tests

- Manifest pins: escrow reports zero drops + emitted abort theorems; the ADT vault example still reports its drop (lane gate).
- Unit tests: the escrow ctx surface (structure, transition signature, guard clause, abort binder, applyOp threading) and a flat dotted-auth fixture binding an imported state field.

Docs: `framework-support.md` footnote + the `<account>.pubkey` accessor section in the DSL reference.

Closes #328


abishekk92:
CI caught a real regression the local suite could not: the release gate's zero-sorry sweep over the example baselines. Routing escrow's authorization clauses through ctx had flipped its two cover witness proofs and the liveness auto-proof to `sorry`. Fixed in 262dc6b rather than worked around: the liveness split-script is guard-agnostic (ctx binder is safe — gate removed), and covers now construct an all-`pk` `ActionCtx` witness, matching the witness machinery's existing default-pubkey convention. Escrow is back to zero `sorry` with the proofs now over the faithful model — main proved them over the weakened guard. `lake build` and `scripts/release-gate.sh` pass locally.

PR #351, opened 2026-07-23
fix(proptest): three harness-correctness gaps found dogfooding a vault spec

saicharanpogul:
Dogfooding a vault spec against the proptest backend surfaced three distinct defects in the generated harness. Each is confirmed with a reproduction that `qedgen check` accepts (0 errors) but the emitted `proptest.rs` mishandles, plus a regression test. All local gates green (`cargo test`, `fmt`, `clippy --all-targets -D warnings`, `check --regen-drift`, `check-readme-drift.sh`, `release-gate.sh`, `codegen_smoke --ignored`).

## 1. Undeclared init pre-state dropped from the lifecycle → compile break

An init handler `State.Uninitialized -> State.Active` sets `pre_status = "Uninitialized"`, but the account lifecycle is built purely from declared ADT variant names (`adapt.rs`). When the State ADT doesn't declare `Uninitialized`/`Empty` — a valid *implicit* init pre-state the state lints already accept — the generated `Status` / `Lifecycle` enums lacked the variant the transition fns and sequence transition table reference:

```
error[E0599]: no variant or associated item named `Uninitialized` found for enum `Status`
error[E0599]: no variant or associated item named `Uninitialized` found for enum `Lifecycle`
```

The sequence also seeded the wrong initial state (`.first()` = the first *declared* variant), running non-init handlers before `init`.

**Fix:** inject a referenced-but-undeclared sentinel into the account lifecycle — a pre-state sentinel becomes the initial state, a post-only one terminal. Specs that already declare the sentinel are unchanged (dedup).

## 2. A field bound leaked onto unrelated fields → unreachable values

`arb_state` propagated the tightest `state.F <= CONST` bound to **every** unbounded numeric field (so relational properties like `V >= C + I` stay satisfiable in random states). But a field in no such property — e.g. a type discriminant — got capped too, so `token_kind = 2` became unreachable and proptest silently never explored it (`token_kind in 0u8..=1u8`).

**Fix:** propagate a bound only to fields that co-occur with an explicitly-bounded field in some property expression. Unrelated fields keep their full domain.

## 3. A ghost accumulator overflowed under full-domain `arb_op` amounts

A ghost aggregate (`total := total + amount`) renders with plain arithmetic and has no reject path (unlike a checked state field). The sequence harness fed it `arb_op` params up to `u64::MAX`, and its header runs in debug (`cargo test`), where the accumulation panics:

```
thread 'state_machine_sequence' panicked: attempt to add with overflow
```

In release it wraps instead, spuriously violating a conservation property.

**Fix:** bound an `arb_op` param to `type_max / SEQ_LEN` **only** when a ghost update on its handler reads it — so `SEQ_LEN` additions stay in-domain. Params feeding only checked state fields (which reject on overflow) keep the full domain, as the dedicated per-handler overflow tests still need.

## Not reproducible on current `main` (from the same report)

Two further items were investigated and are **not** bugs on `main`, so they're not touched here:
- The generated `Cargo.toml` already carries the `proptest` dev-dependency unconditionally (regression-tested in `codegen_mir`).
- `arb_state` already samples the full u8 domain outside the bound-leak fixed in (2).

## Blast radius

Bundled examples are unchanged (they declare `Uninitialized`, have no bound leak, and — lending/multisig — no ghosts). Only the product-state ghost snapshot regenerates, because its deposit amount feeds a ghost. Three regression tests added (one per fix).

### DIFF
--- a/crates/qedgen/src/codegen/proptest_gen_mir.rs
+++ b/crates/qedgen/src/codegen/proptest_gen_mir.rs
@@ -863,12 +863,38 @@ fn emit_account_section(
     // so that relational invariants like `V >= C_tot + I` have valid input ranges.
     let mut field_bounds = extract_field_upper_bounds(properties, &spec.constants);
     if !field_bounds.is_empty() {
-        // Find the tightest bound and apply it to all unbounded numeric fields
-        // of the same type. This ensures relational properties hold in random states.
+        // A relational property like `V >= C_tot + I` is unsatisfiable in
+        // random states unless the fields it relates share a compatible
+        // upper bound: if `V <= MAX` but `C_tot` samples the full u64
+        // domain, `V >= C_tot + I` almost never holds and the harness
+        // rejects everything. So propagate the tightest explicit bound —
+        // but ONLY to fields that co-occur with an explicitly-bounded
+        // field in some property expression. Blanket-capping every numeric
+        // field (the old behavior) silently shrank the domain of fields in
+        // NO such property — e.g. a type discriminant with its own wide
+        // range became unreachable above the borrowed bound.
+        let explicit: std::collections::HashSet<String> = field_bounds.keys().cloned().collect();
         let min_bound = field_bounds.values().min_by_key(|v| v.len()).cloned();
         if let Some(ref bound) = min_bound {
+            let mut related: std::collections::HashSet<String> = std::collections::HashSet::new();
+            for prop in properties {
+                if prop.expression.is_none() {
+                    continue;
+                }
+                let mentioned: Vec<&str> = mutable_fields
+                    .iter()
+                    .filter(|(f, _)| property_references_field(prop, f))
+                    .map(|(f, _)| f.as_str())
+                    .collect();
+                if mentioned.iter().any(|f| explicit.contains(*f)) {
+                    related.extend(mentioned.iter().map(|f| f.to_string()));
+                }
+            }
             for (fname, ftype) in mutable_fields {
-                if ftype.as_str() != "Pubkey" && !field_bounds.contains_key(fname.as_str()) {
+                if ftype.as_str() != "Pubkey"
+                    && related.contains(fname.as_str())
+                    && !field_bounds.contains_key(fname.as_str())
+                {
                     field_bounds.insert(fname.to_string(), bound.clone());
                 }
             }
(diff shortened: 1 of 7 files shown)


abishekk92 (changes requested):
I found three correctness/coverage issues that should be addressed before merge:

1. [P1] Ghost overflow protection is not expression-aware. seq_param_bound caps a referenced parameter at type_max / SEQ_LEN regardless of how the parameter is used. An accepted update such as volume += amount * 21 still emits plain debug arithmetic while allowing amount <= u64::MAX / 20, so it can overflow on the first operation. The bound must account for the update expression, or the generated ghost arithmetic needs safe semantics.

2. [P2] Non-accumulator ghost assignments lose sequence coverage. The detector treats any ghost update mentioning a parameter as an accumulator. For last_amount := amount, generation restricts amount to u64::MAX / 20 even though assignment cannot accumulate overflow, making the upper 95% of the operation domain unreachable. Detection should require a self-referential additive accumulator rather than merely a parameter reference.

3. [P2] Bounds still leak between unrelated property groups. The propagation code computes one global minimum bound. With separate groups a <= 10; a >= b and c <= 1000; c >= d, generation produces d in 0..=10 instead of borrowing the bound from c. Propagation should be per property/connected component rather than using the global minimum.

I reproduced all three against the PR head. The 27 nearby proptest generator tests, including the new regressions, pass.

saicharanpogul:
Thanks — all three reproduced here too, and all three are fixed in `bd2e2f2`. The through-line: I dropped the input-bounding heuristic entirely (which is what issues 1 and 2 were both attacking) in favor of the "safe ghost semantics" you offered as the alternative.

**1 (P1) — expression-aware overflow.** Replaced `seq_param_bound` with **saturating** ghost arithmetic. Ghost updates now render `saturating_add`/`sub`/`mul`, recursing into nested terms, so your `volume += amount * 21` becomes:

```rust
s.volume = (s.volume).saturating_add((amount).saturating_mul(21));
```

It clamps at the type bound for *any* update expression — no coefficient analysis, no panic. I used saturating rather than wrapping deliberately: wrapping drops the aggregate *below* a co-tracked value and breaks `ghost >= field`, whereas saturating keeps it at the max, preserving the monotone conservation property. (`div`/`rem` have no saturating form and can't overflow, but a zero divisor would panic, so they clamp to 0.)

**2 (P2) — non-accumulator coverage.** With saturating semantics there's no input bounding at all, so the accumulator-detection heuristic is gone. `last_amount := amount` — and every other param — keeps its full `0u64..=u64::MAX` domain. The whole class disappears rather than needing a tighter detector.

**3 (P2) — cross-group bound leak.** Propagation is now per connected-component: fields are unioned by co-occurrence in a property expression, and each component applies its own tightest bound (compared numerically, not by string length). Your repro now yields `d in 0u64..=1000u64` (borrowing `c`'s bound), not `0..=10`. Regression test `bound_propagation_is_per_component_not_global_minimum` asserts exactly that shape.

Regenerated the ghost-bearing snapshots (saturating render in both the proptest and Kani lanes — the Kani change drops a ghost-overflow false positive). Bundled examples are untouched. All local gates green (test, fmt, clippy -D warnings, regen-drift, readme-drift, release-gate, codegen_smoke --ignored).

abishekk92:
Thanks. The three fixes look good and CI is green.

One question: should `SaturatingEffect` preserve the existing total arithmetic semantics? It emits `x % 0 = 0`, while Lean and the widened renderer use `x % 0 = x`. For signed ghosts, `MIN / -1` also becomes `0` instead of saturating. Are these behaviors intentional?

saicharanpogul:
Good catch — not intentional, now fixed in `608d74f`. The intent is for `SaturatingEffect` to preserve the existing total-arithmetic semantics, so I aligned div/mod with `render_widened_term` / Lean:

- **`x % 0`**: was `unwrap_or(0)`, now `checked_rem(r).unwrap_or(l)` → `x % 0 = x`, matching `render_widened_term` (tree_render.rs:1210) and Lean.
- **`x / 0`**: unchanged at `checked_div(r).unwrap_or(0)` → `x / 0 = 0`, which already matched `render_widened_term` (:1206).
- **Signed `MIN / -1`**: resolves to `0` — this is the same result `render_widened_term`'s `checked_div(...).unwrap_or(0)` produces, i.e. the established total-division convention, not a separate saturating rule. Saturation is reserved for `+`/`-`/`*` (the ops that actually overflow-panic in a monotone aggregate); `div`/`rem` have no saturating form, so they follow the total convention rather than inventing one. So the ghost model stays consistent with the proofs.

Added `saturating_effect_saturates_addsubmul_and_keeps_total_div_mod` pinning all five renderings. No fixture ghost uses div/mod, so no snapshots move; full suite + gates green.

abishekk92 (approved):
Re-reviewed through 608d74f. The three requested correctness/coverage fixes are addressed: ghost arithmetic is expression-safe without shrinking arb_op domains, bound propagation is scoped per connected property component, and implicit lifecycle sentinels are injected consistently. The div/mod follow-up now matches the established total-arithmetic convention. Focused generator/renderer tests pass locally and all required CI checks are green. LGTM.

PR #373, opened 2026-07-27
feat(qedgen): add Parallax integration tests and executed reproducers

abishekk92:
## Summary

- generate Parallax integration test scaffolds for supported program shapes
- generate and execute missing-signer and lifecycle replay reproducers before promoting probe findings
- pin the Parallax dependency, add compile/runtime CI gates, and document pin maintenance
- harden checked-arithmetic error handling and correct the Quasar compile claim

## Testing

- cargo test
- cargo clippy -- -D warnings
- npm test
- cargo test -p qedgen-solana-skills --test parallax_integration_gate -- --ignored

## Known review blockers

This is intentionally a draft. Review identified correctness issues that should be resolved before merge:

- bind the selected program ID and deploy artifact to the same exact crate in multi-program workspaces
- synthesize checked-arithmetic error variants before the empty-error early return and handle overflow and underflow defaults independently
- distinguish expected attack rejection from runtime, setup, and unrelated harness failures instead of classifying nearly every test failure as not reproduced

The existing uncommitted documentation files in the local checkout are not included in this PR.


abishekk92 (commented):
Blocking review: do not merge this PR yet. The current diff can independently select a declare_id from one crate and the first deploy artifact from another, so a multi-program workspace can execute the wrong binary. execute_parallax_harness also maps almost every non-compiler cargo test failure to NotReproduced, including setup or program-load failures unrelated to the guard. Finally, checked-arithmetic variants are resolved only after emit_errors returns early for an empty declared Error set. Select one concrete crate and derive both its address and named artifact from it; recognize only an explicit machine-readable attack-rejection signal; and gate errors.rs plus pub mod errors on emitted_error_variants. These three blockers are still present at the current stack tip.

PR #374, opened 2026-07-27
feat(verify): compile generated program scaffolds

### LINKED ISSUE #364: G22: verify has no backend that compiles the generated program crate — non-compiling codegen reaches the user with check reporting 0 errors
## Summary

Nothing in the user-facing loop ever compiles the generated program crate.

- `qedgen check` is spec-level lint only.
- `qedgen verify` compiles the HARNESSES, and does so in isolated crates on
  purpose (`crates/qedgen/src/verify/mod.rs:392-397`: "Run in an isolated
  crate: the harness is framework-neutral, and tying it to the program
  package lets unrelated scaffold compile errors block Kani").

So when codegen emits Rust that does not compile, the user finds out from
`cargo build`, with no qedgen diagnostic anywhere and often after `check`
printed `0 error(s)`.

The maintainer-side gate that does catch this class
(`crates/qedgen/tests/generated_artifact_gate.rs`) runs only over the bundled
examples in CI. It is corpus-bound: it catches a defect only if some bundled
spec happens to exercise the shape. It cannot see a user's spec at all.

This class keeps recurring — several closed issues in this tracker are
individual instances of "generated Rust does not compile" found by a human,
plus one more found this week.

## Repro

Any spec whose generated program references an undeclared error variant
(a companion issue covers that specific defect):

```
$ qedgen check --spec e.qedspec
0 error(s), 1 warning(s), 1 info      # the warning is unrelated
Generated 6 files in ./programs
$ grep -n 'Error::UndeclaredRequiresErr' programs/src/guards.rs
$ grep -c UndeclaredRequiresErr programs/src/errors.rs
0
```

Observed: exit 0 from the loop; the defect is only visible to `cargo build`.
Expected: qedgen tells the user its own output does not compile.

Confirming the absence of any such step:
`grep -rn 'Command::new("cargo")' crates/qedgen/src` returns proptest, Kani,
Miri, the Crucible fuzz build, and probe reproducers — never the generated
program crate.

## Proposed fix

Add a `scaffold` backend to `verify`:

- run `cargo check --tests` in the generated program crate;
- report it as a `BackendReport` like proptest/Kani, with `skipped` +
  reason when the crate is absent or cargo is unavailable;
- record it in the obligation manifest so `verify --strict` can gate on it.

Cheap to implement: the crate already exists on disk after `codegen`, and
the `BackendReport` plumbing, failure summarization, and manifest recording
are all in place.

Optionally the same check can run as a post-`codegen` opt-in
(`codegen --check-compiles`) so the loop closes at generation time.

## Gate

A CLI test that runs codegen, edits the generated `guards.rs` to reference an
undeclared enum variant, runs `verify`, and asserts the `scaffold` backend
reports `failed` with the rustc error attached.

## Why this matters

It converts the whole "codegen emits non-compiling Rust" family from
user-discovered to tool-discovered, and does it property-wise rather than
corpus-wise: it works on the user's own spec, so it does not require a
bundled example per shape.

abishekk92:
## Summary

- add a first-class scaffold verify backend that runs cargo check --tests in the selected program crate
- auto-enable scaffold for flagless verify --program while keeping explicit backend selection exact
- report bounded rustc diagnostics, gate enabled skips under --strict, and retain source-bound evidence without authorizing stamp
- document the workflow and its buildability-only trust boundary

The issue proposed adding scaffold to the semantic obligation manifest. This PR keeps it in verify evidence instead: compilation establishes buildability, not a qedspec semantic obligation, so it must not affect obligation coverage or stamp eligibility.

## Tests

- cargo test
- cargo clippy -- -D warnings
- npm test
- bash scripts/check-readme-drift.sh
- cargo test -p qedgen-solana-skills --test scaffold_verify_cli -- --nocapture

Closes #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,15 @@ qedgen init --name my_program --spec my_program.qedspec --target anchor
 qedgen check
 qedgen codegen --all
 
-# 3. Fill generated Rust handler TODOs, then run backend verification
-qedgen verify
+# 3. Fill generated Rust handler TODOs, then verify the program crate
+qedgen verify --program ./programs/my_program
 ```
+
+When `--program <crate>` is supplied without explicit backend flags, verify
+also runs the `scaffold` backend (`cargo check --tests`) so generated Rust
+compile failures surface inside QEDGen. Use `--scaffold` to combine that check
+with an explicitly selected backend. A scaffold pass proves buildability, not
+spec conformance, and cannot authorize `qedgen stamp`.
 
 ### Stuck? File feedback
 
--- a/crates/qedgen/src/cli.rs
+++ b/crates/qedgen/src/cli.rs
@@ -750,9 +750,11 @@ pub(crate) enum Commands {
 
     /// Run the generated harnesses against the generated implementation.
     ///
-    /// `check` validates the spec; `verify` validates the code the spec
-    /// produced. Default (no flags) runs every backend whose artifact is
-    /// present on disk. Use --proptest/--kani/--lean to target one backend.
+    /// `check` validates the spec; `verify` runs proof backends and can
+    /// compile a selected program crate. Default (no flags) runs every
+    /// backend whose artifact is present on disk. With `--program`, it also
+    /// runs `cargo check --tests` to establish buildability, not semantic
+    /// conformance. Use --proptest/--kani/--lean to target one backend.
     Verify {
         /// Path to the spec file (.qedspec). Optional — falls back to the
         /// `spec` field in the nearest `.qed/config.json` discovered by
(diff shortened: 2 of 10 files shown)


abishekk92 (commented):
Review verdict: supersede this PR with #377 rather than merge two implementations of the scaffold backend. If this implementation is retained, fix the implicit --program path before merge: the early-return lanes can exit after --check-upstream or --probe-repros before the no-explicit-flags logic auto-enables scaffold. Centralize one backend-request predicate and reuse it at every early return.

abishekk92:
Superseded by #377, which included the scaffold verification backend plus the follow-up fixes and has now landed through #373 into main.

PR #375, opened 2026-07-28
fix(proptest): sequence harness respects each property's preserved_by scope

saicharanpogul:
## The bug

`state_machine_sequence` asserted **every** property after **every** successful op, ignoring each property's `preserved_by` scope. A property `preserved_by [deposit]` is an obligation on `deposit` alone — another handler may legitimately leave it false — so asserting it after an unrelated op produces a **false counterexample on a correct spec**.

## Reproduction (generic ledger)

```
type Ledger | Active of { total : U64, locked : U64 } | Closed

handler deposit (amount : U64) : Ledger.Active -> Ledger.Active {
  requires amount > 0 else InvalidAmount
  effect { total += amount }
}
handler lock (amount : U64) : Ledger.Active -> Ledger.Active {
  requires amount > 0 else InvalidAmount
  effect { locked := state.locked + amount }
}
property total_ge_locked :
  state.total >= state.locked
  preserved_by [deposit]
```

`lock` grows `locked` and is **not** obligated to preserve `total_ge_locked`. The old harness asserted it after `lock` anyway:

```
Test failed: total_ge_locked violated after op Lock(1)
```

— a false failure on a correct spec.

## The fix

Two parts, applied to both `emit_sequence_test_for` and the product `emit_product_sequence_test`:

1. **Scope** — assert a property only after handlers in its `preserved_by` set. `preserved_by all` expands to the full handler list, so it still asserts after every op.
2. **Conditional** — `preserved_by` is `pre → post`: a handler must keep a property that held *before* it, but need not *restore* one a prior op already broke (`[lock(big), deposit(1)]` leaves it false before `deposit`). The loop snapshots the pre-state (`let pre = s;`; `State`/`ProductState` are `Copy`) and emits:

```rust
if matches!(op, Op::Deposit(..)) && total_ge_locked(&pre) {
    prop_assert!(total_ge_locked(&s), "...");
}
```

For `preserved_by all` the match guard is omitted (every op qualifies), leaving just the pre-state check.

## Still finds real bugs

The change is scope-aware, not a blanket weakening. A `preserved_by all` invariant a handler genuinely breaks still yields its minimal counterexample — verified with a generic `backing >= claim` spec whose `spend` handler drops `backing`:

```
backing_ge_claim violated after op Spend(1)
minimal failing input: [Fund(..), Spend(..)]
```

## Validation

- New regression test `sequence_asserts_scoped_property_only_after_its_preserving_handler`.
- Regenerated `lending` / `multisig` example proptests + the product-ghost snapshot (the change is the pre-snapshot plus the scoped guard). The regenerated multisig example proptest passes all 23 tests at 2000 cases.
- All gates green: `cargo test`, `fmt --check`, `clippy --all-targets -D warnings`, `check --regen-drift`, `check-readme-drift.sh`, `release-gate.sh`, `codegen_smoke --ignored`.

### DIFF
--- a/crates/qedgen/src/codegen/proptest_gen_mir.rs
+++ b/crates/qedgen/src/codegen/proptest_gen_mir.rs
@@ -1991,6 +1991,34 @@ fn emit_sequence_test_for(
         out.push_str("            }\n");
     }
 
+    // Properties this section checks in the sequence: those with at least
+    // one obligated handler here. A pre-state snapshot lets each check
+    // assert the `preserved_by` obligation faithfully — `pre → post`, not
+    // just `post` (see the emission comment below).
+    let seq_props: Vec<(&&ParsedProperty, Vec<String>)> = all_props
+        .iter()
+        .filter_map(|prop| {
+            let pats: Vec<String> = handlers
+                .iter()
+                .filter(|h| prop.preserved_by.iter().any(|n| n == &h.name))
+                .map(|h| {
+                    let pascal = crate::codegen_shared::to_pascal_case(&h.name);
+                    if model_params(h).next().is_some() {
+                        format!("Op::{}(..)", pascal)
+                    } else {
+                        format!("Op::{}", pascal)
+                    }
+                })
+                .collect();
+            (!pats.is_empty()).then_some((prop, pats))
+        })
+        .collect();
+
+    // Snapshot the pre-state (State is `Copy`) before the transition so the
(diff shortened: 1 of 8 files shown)


abishekk92 (approved):
Approved. The generated sequence checks now match preserved_by semantics: only obligated handlers are checked, and preservation is emitted as pre implies post. The single-account and product-state paths are both covered, CI is green, and I found no blocking issue in this diff.

PR #376, opened 2026-07-28
fix(proptest): arb_state satisfies conservation invariants by construction

saicharanpogul:
## The bug

Preservation tests filter random `arb_state()` states through `prop_assume!(invariant(&pre))`. For a tight relational invariant — a conservation law like `total == a + b`, or several stacked — independent full-range field generation almost never satisfies the conjunction, so `max_global_rejects` is exhausted and the test **aborts** (`Too many global rejects`) having validated nothing. A genuine handler bug in that state space is then invisible.

## Reproduction (generic)

```
property conserved : state.total == state.a + state.b preserved_by [mint]
```

**Before:** `mint_preserves_conserved` → `Test aborted: Too many global rejects`.
**After:** `arb_state` emits `let total = (a).saturating_add(b);` → test runs, passes in 0.03s.

## The fix

`arb_state` repairs the generated fields so the common conservation shape `field (>=|<=|==) <sum of other fields (+ consts)>` holds by construction. It recovers those constraints from the spec's unary properties and invariants (both operand orders, `and`-split conjunctions), then emits `let field = field.max(rhs)` / `.min(rhs)` / `= rhs` shadows **in dependency order** (a field feeding another's RHS is repaired first):

```rust
// property total >= reserved + pending ; pending >= queued
let pending = pending.max(queued);
let reserved = reserved.max(locked);
let total = total.max((reserved).saturating_add(pending));
```

Only `arb_state` is repaired; `arb_boundary_state` (guard-rejection tests, which *want* invalid states) is untouched.

## Why it's safe

Correctness rests on the preservation tests **keeping their `prop_assume`** as the safety net — the repair only lowers the reject rate, so any shape it doesn't recognize (subscripts, `-`/`*` on the RHS, mixed `>=`/`<=` on one field, cyclic dependencies) reject-samples exactly as before. **No regression.**

## It does not hide bugs — it un-hides them

A handler that genuinely breaks the invariant still fails. A `mint` that grows `total` but not `a`/`b`:

```
Test failed: conserved must hold after mint
```

Before, this bug was invisible (the test aborted on rejects, validating nothing). The repaired pre-state now lets the test actually **run** and catch it.

## Validation

- Regression tests `arb_state_repairs_conservation_invariants_in_dependency_order`, `arb_state_repair_falls_back_on_unsupported_shape`.
- Bundled `lending` (`total_deposits >= total_borrows`) and `multisig` (`member_count >= approval_count + rejection_count`) pick up one repair line each; the multisig example proptest still passes all 23 tests (and faster — fewer rejects).
- All gates green: `cargo test`, `fmt --check`, `clippy --all-targets -D warnings`, `check --regen-drift`, `check-readme-drift.sh`, `release-gate.sh`, `codegen_smoke --ignored`.

Independent of the sequence-scope fix in the sibling PR #375 (different functions; both target `main`).

### DIFF
--- a/crates/qedgen/src/codegen/proptest_gen_mir.rs
+++ b/crates/qedgen/src/codegen/proptest_gen_mir.rs
@@ -1101,16 +1101,16 @@ struct FieldConstraint {
 /// return its name.
 fn tree_state_field(tree: &crate::mir::ExprTree) -> Option<String> {
     use crate::mir::expr_tree::{BindingKind, TreeSeg};
-    if let crate::mir::ExprTree::Path(p) = tree {
-        if matches!(p.binding, BindingKind::StateField | BindingKind::Ghost)
-            && p.segments.len() == 1
-        {
-            if let TreeSeg::Field(f) = &p.segments[0] {
-                return Some(f.clone());
-            }
-        }
+    let crate::mir::ExprTree::Path(p) = tree else {
+        return None;
+    };
+    if !matches!(p.binding, BindingKind::StateField | BindingKind::Ghost) {
+        return None;
     }
-    None
+    match p.segments.as_slice() {
+        [TreeSeg::Field(f)] => Some(f.clone()),
+        _ => None,
+    }
 }
 
 /// Render `tree` as a saturating-`+` chain over bare field names / int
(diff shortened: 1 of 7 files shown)


abishekk92 (approved):
Approved. The repair is deliberately best-effort and keeps prop_assume as the correctness safety net, while dependency ordering handles the supported conservation shapes. CI is green. Merge #375 first, then rebase and regenerate this PR because both touch the same emitter and generated snapshots.

Issue #372, opened 2026-07-27
B17: codegen --target quasar emits a program that does not compile — quasar-lang uses solana_address::Address, not Pubkey


abishekk92:
## Summary

`codegen --target quasar` emits a program that does not compile. Quasar is listed as a fully supported target and three bundled examples target it, yet no Quasar artifact in this repo has ever been built.

## Evidence

Generated from the bundled multisig spec against a clean tempdir, with only the `qedgen-macros` path patched:

```
$ qedgen codegen --spec multisig.qedspec --target quasar --output-dir programs
Generated 14 files in .../programs
$ cargo check
error[E0412]: cannot find type `Pubkey` in this scope
 --> src/state.rs:5:27
error[E0425]: cannot find type `Pubkey` in this scope
 --> src/lib.rs:58:24
...
error: could not compile `multisig` (lib) due to 13 previous errors
```

## Root cause

`quasar-lang` 0.0.0 **is** published and resolves (1743 downloads). It is a substantive `#![no_std]` framework, and its addresses are `solana_address::Address` — it defines no `Pubkey` at all:

```
$ grep -rn "Pubkey" ~/.cargo/registry/src/*/quasar-lang-0.0.0/src/lib.rs
(no matches)
$ grep -n "keys_eq" ~/.cargo/registry/src/*/quasar-lang-0.0.0/src/lib.rs
132: pub fn keys_eq(a: &solana_address::Address, b: &solana_address::Address) -> bool {
```

`map_type_with_context` maps `Pubkey` identically for Anchor and Quasar:

```rust
// codegen_shared/types.rs:703
"Pubkey" => match context {
    TypeMapContext::Anchor | TypeMapContext::Quasar => "Pubkey",
```

So the Quasar target inherits Anchor's type name for a framework that uses a different one.

## Why it survived

No gate compiles a Quasar artifact. The snapshot suites (`tests/*_snapshot.rs`) compare generated TEXT, which is stable and wrong in exactly the same way every run. The executable artifact gate (`tests/generated_artifact_gate.rs`, #294) is corpus-bound to Anchor examples. `check --regen-drift` regenerates and diffs, but never builds.

This is the same class as #364: nothing in the user loop compiles the generated program crate.

## Suggested fix

Split the Quasar arm of the `Pubkey` mapping to Quasar's address type, then add a Quasar example to the artifact gate so the target cannot regress to non-compiling again. The second half is the part that matters — without it this recurs silently.

[...]

Issue #363, opened 2026-07-27
G21: error-variant declaration and use are two lists — requires/abort names emit undeclared variants and check reports 0 errors


abishekk92:
## Summary

Error-variant DECLARATION and error-variant USE are still derived
independently. A recent fix made checked arithmetic (`+=` / `-=`) share one
resolver between the `errors.rs` enum and the `ok_or(<Prog>Error::…)?` site.
Every other place a spec can name an error variant is still unreconciled:

- `requires <expr> else <Variant>` — emitted into `guards.rs`, never
  declared, never linted.
- a `match` arm's `abort <Variant>` — same (it desugars to a synthetic
  `requires` with `error_name`).
- an effect's `else <Variant>` — now auto-declared by the arithmetic fix, but
  the lint still tells the user it "won't compile".
- `integration_test::authorization_error` — deliberately under-approximates
  the enum because it has no way to ask what will be declared.

Result: `qedgen check` reports `0 error(s)` on a spec whose generated program
does not compile.

## Repro

Spec (Anchor target, no `program_id` needed):

```
type Error
  | InvalidAmount

handler settle (amount : U64) : State.Active -> State.Active {
  auth owner
  accounts { owner : signer, writable
             vault : writable }
  requires amount > 0 else UndeclaredRequiresErr
  effect { total := 0 }
}
```

```
$ qedgen check --spec e.qedspec --json | jq '[.findings[].rule]'
["preserved_by_all_potential_violation","pubkey_state_field_unsupported"]
# summary: 0 error(s) — no unknown_error_variant

$ qedgen codegen --spec e.qedspec
$ grep -A4 'pub enum' programs/src/errors.rs
pub enum ErrVar5Error {
    InvalidAmount = 0,
    InvalidLifecycle = 1,
}
$ grep -n UndeclaredRequiresErr programs/src/guards.rs
27:        return Err(ErrVar5Error::UndeclaredRequiresErr.into());
```

Observed: enum has no such variant, program does not compile, check is
silent.
Expected: either the variant is declared (as the arithmetic path now does) or
`check` reports it.

The `abort` form reproduces identically:

```
  match
    | state.total >= amount => effect { total := 0 }
    | _                     => abort AbortTypoVariant
```
→ `guards.rs:45: return Err(<Prog>Error::AbortTypoVariant.into());`, enum
unchanged, `check` silent.

The converse drift, same spec shape with `total += amount else TypoVariant`:
[...]

Issue #294, opened 2026-07-20
ci(codegen): gate generated artifacts by compilation and execution


abishekk92:
## Context

PR #263 fixes five codegen-quality gaps found by dogfooding a Fortel-vault-shaped spec: Bool fixture defaults, comparison-valued Kani assertions, PDA-signed token CPIs, compound guard fixtures, and lifecycle-driven PDA initialization.

These are not five unrelated mistakes. Together they expose a process gap: we validate much of codegen as stable generated text, while users consume it as compiled and executed software.

This issue depends on #263: land or otherwise resolve #263 first, preserving its reproductions and regression coverage. Then use those cases to harden the repository-wide quality gates rather than treating the fixes as isolated regressions.

## How the gaps escaped

- The bundled Anchor examples are excluded from the root Cargo workspace, so root `cargo test` does not compile or run their generated unit tests.
- The Anchor scaffold smoke compiles escrow and multisig, but not lending; it does not compile/run generated unit tests or compile the generated Kani artifact, and only escrow's proptest is run.
- Snapshot and regen-drift gates prove output stability, not that every emitted artifact compiles or behaves correctly.
- Existing fixtures cover features mostly in isolation. The corpus lacked the interaction of Bool state, comparison-valued effects, lifecycle init, PDA CPI authority, and compound cross-field guards.
- Runtime-only semantics such as PDA signer propagation are invisible to `cargo check`.
- Several code paths bypass canonical typed information: a numeric catch-all for defaults, open-coded expression concatenation, account-name heuristics, and independently derived account attributes/CPI signer behavior.
- The guard solver represented unsupported constraints as partial success and emitted assertions over guessed fixtures.
- The ownership boundary is ambiguous: project guidance says CPI authority/PDA creation may remain agent-filled, while codegen partially mechanizes both. Once codegen emits these operations, it must either own their complete semantics or emit an explicit fill site.

Review of #263 also found the same pattern inside the fix: stale-base compile failure, a fallback fixture that could panic on division/modulo by zero, and unchecked `U128`→`i128` solver arithmetic. This reinforces the need for gates stronger than happy-path emitter tests.

### P0 — executable generated-artifact gate

Add a CI test that, from clean temporary directories:

1. Generates representative Anchor projects with all applicable artifacts.
2. Compiles the generated scaffold, unit tests, proptest harness, and Kani harness (Kani proof execution may remain in its dedicated workflow, but ordinary Rust compilation must be gated).
[...]

PR #377, opened 2026-07-28
feat(verify): add a scaffold backend that compiles the generated program crate

### LINKED ISSUE #364: G22: verify has no backend that compiles the generated program crate — non-compiling codegen reaches the user with check reporting 0 errors

abishekk92:
Closes #364.

`verify --scaffold` runs `cargo check --tests` over the generated program
crate. Nothing in the user-facing loop did that before: `check` is
spec-level lint, and `verify` builds the Kani and proptest harnesses in
isolated crates on purpose, so a codegen defect reached the user as a red
`cargo build` with no qedgen diagnostic, often right after `check` printed
`0 error(s)`. The maintainer-side gate that catches this class
(`generated_artifact_gate`, #294) is corpus-bound to bundled Anchor
examples and cannot see a user's spec at all.

## Classification is the design

A failed `cargo check` is not evidence of a codegen defect on its own. The
generated `Cargo.toml` pins `qedgen-macros` to a git tag at the current
crate version, which does not exist upstream until release time. A
developer build therefore fails to resolve for a reason unrelated to the
spec. Reporting that as a failure would fire constantly, and a backend
that cries wolf gets ignored, which would leave the class as open as it is
now.

So `verify/scaffold.rs` splits on where the failure was reported.
`cargo check --tests --message-format=json` emits rustc diagnostics on
stdout carrying the `manifest_path` of the package they belong to, while
cargo's own failures produce plain stderr and no such objects.

| Signal | Outcome | Backend status |
|---|---|---|
| exit 0 | `Compiled` | passed |
| rustc error with this crate's `manifest_path` | `TypeErrors` | failed |
| rustc error from a dependency | `Unresolved` | skipped, with reason |
| non-zero exit, no rustc diagnostics | `Unresolved` | skipped, with reason |
| no `Cargo.toml` | `NotApplicable` | skipped, with reason |

## Verified against the real defect

Generated an Anchor program from a spec carrying #363's undeclared error
variant (`requires amount > 0 else UndeclaredRequiresErr`):

```
[FAIL] scaffold   (19981 ms)
  error[E0599]: no variant or associated item named `UndeclaredRequiresErr`
                found for enum `errors::ErrVarError` in the current scope
    --> src/guards.rs:27:33
   ::: src/errors.rs:5:1
     5 | pub enum ErrVarError {
```

Exit 1. That is a spec `qedgen check` reports `0 error(s)` on today.

## Two things the build surfaced

**`cargo check` rewrites `Cargo.lock`.** The codegen snapshot suite caught
it: one run repinned all three committed example locks to a different
resolution. A check that mutates what it checks is a worse bug than the one
it finds. Fixed with `--locked`, but only when a lock already exists, since
`--locked` on a freshly generated crate fails outright.

**A scaffold pass must not authorize `stamp`.** The backend compiles the
real program crate, so it looks implementation-bound. It is not. Letting it
count would let a green `cargo check` license `#[qed(verified)]`. Explicit
arm in `evidence.rs` rather than the catch-all, so the decision is recorded
where the next backend author reads it.

## Dropped from the issue

The issue proposed recording the result in the obligation manifest so
`verify --strict` could gate on it. The manifest is one entry per
spec-derived obligation (`{backend, kind, scope, key, status}`), built as a
pure in-memory render so it can never be stale-file drift. "The crate
compiles" has no obligation key and is a filesystem fact. It is also
unnecessary: `run.rs` already exits 1 on any failed backend.

## Reach

`codegen` runs the same check over what it just wrote, so the defect is
caught in the command that produced it. It runs only once the dependency
tree resolves (a `Cargo.lock` in the crate or its parent), so a first
generation never pays a cold `anchor-lang` build and prints one line saying
the check was deferred. It never changes the exit code: codegen's contract
is to write files, and a brownfield crate that fails to build for unrelated
reasons must not make codegen unusable. Opt out with
`--no-check-compiles`.

## Tests

- 13 unit tests in `verify/scaffold.rs` over the classifier and argument
  construction, including the dependency-error and cargo-level cases and
  the `--locked` regression.
- `tests/scaffold_backend_gate.rs`: 5 fast tests using a dependency-free
  crate, plus one `#[ignore]` end-to-end case that generates a real Anchor
  program carrying the #363 defect. Wired into CI next to the other
  compile-heavy gates.
- `codegen_snapshot.rs` passes `--no-check-compiles`: those fixtures carry
  a `programs/Cargo.lock`, and the suite reads generated text only.

Full suite green (39 suites, 1634 tests). clippy and rustfmt clean.

## Notes

Stacked on #373, which this depends on for the CI job adjacency and the
`framework-support.md` footnote ordering. Retarget to `main` once #373
merges.

`docs/framework-support.md` gains a row for the new backend. Anchor is
gated, Quasar fails today (#372), and Pinocchio is marked "compiles,
ungated" after generating one and running the check by hand rather than
asserting it from the code.

This makes #372 mechanically visible: the three bundled Quasar examples
will fail `--scaffold` until the `Pubkey` mapping is fixed. That is the
correct reading, not a regression.


abishekk92 (commented):
Blocking review. This is the scaffold implementation I would keep over #374, but it is not merge-ready. Explicit --scaffold is omitted from the early-return predicates around --check-upstream and --probe-repros, so those combinations return without running the requested backend. Include scaffold in one centralized any-backend predicate and test both combinations. Also fold #379 into this PR: generated-artifacts and runtime-journey are currently red here and remain red in #378, while the later child supplies the dependency pins that make the stack green. Every parent should pass against its declared base.

abishekk92 (commented):
Re-review at 98fa980: the blocking issues from my earlier review are resolved. The verify dispatch now uses one centralized any_verify_backend predicate, and black-box regressions prove both --check-upstream --scaffold and --probe-repros --scaffold reach the scaffold backend. The wincode-boundary pins from #379 are folded into this branch without pulling in #378. All refreshed checks are green: test, generated-artifacts, runtime-journey, supply-chain, and Greptile. GitHub does not allow a formal approval on a self-authored PR, but I have no remaining blocking findings on this head.

Issue #371, opened 2026-07-27
F14: the Parallax pin-liveness check is a manual RELEASING.md bullet, not a CI job — its failing condition breaks generated scaffolds for users


abishekk92:
## Summary

`scripts/check-parallax-pin.sh` is wired only into the release checklist, as
a manual step. Its exit-1 condition is a user-facing break, not a release
concern: it fires when the pinned `parallax-svm` revision no longer exists
upstream (force-push, rebase, deleted branch), which means every generated
integration scaffold stops resolving its dev-dependencies. Between releases,
nothing notices.

## Evidence

- `docs/RELEASING.md:17` — the step is numbered "5a", squeezed between 5 and
  6, which is itself a sign the checklist is absorbing work that belongs in a
  gate.
- `grep -n parallax .github/workflows/ci.yml` — CI runs the integration gate
  and the reproducer gate, never the pin check.
- `grep -n parallax scripts/release-gate.sh` — the mechanical release gate
  does not call it either.

The script's exit codes are already designed for automation:

```
0 = pin current, OR behind upstream (expected — pinning is the point),
    OR upstream unreachable (offline / rate limit / CI flake)
1 = the pinned revision no longer exists upstream — the dependency will
    not resolve
2 = could not read the pin from source (the single-source const moved)
```

Nothing about it needs a human.

## Proposed fix

1. Run it from a scheduled CI job (weekly is enough — the failing condition
   is an upstream force-push, not routine churn).
2. Call it from `scripts/release-gate.sh` so the release path is mechanical.
3. Reduce the RELEASING.md step to a pointer at the gate, keeping only the
   pin-bump procedure (which is genuinely manual).

Exit code 2 is worth wiring too: it means the single-source constant was
renamed or moved, so the check silently stopped checking anything.

## Why this matters

Same shape as the previously filed release-checklist drift issue: a check
that exists as prose in RELEASING.md instead of as a gate has a human as its
only owner, and humans skip steps. This one guards a dependency that a
generated artifact points at, so its failure lands in a user's crate.

PR #378, opened 2026-07-28
fix(codegen): map Pubkey to Address for Quasar and gate the target

### LINKED ISSUE #372: B17: codegen --target quasar emits a program that does not compile — quasar-lang uses solana_address::Address, not Pubkey

abishekk92:
Closes #372.

`codegen --target quasar` emitted a program that did not compile. quasar-lang
is `#![no_std]` and defines no `Pubkey`; its prelude re-exports
`solana_address::Address`. `primitive_map` shared one arm between the two
program targets, so every generated Quasar program named a type the
framework does not have.

The bundled `escrow`, `lending`, and `multisig` examples are Quasar crates,
so all three shipped programs that could not build.

## The fix

One arm in `primitive_map`. The issue speculated the fix was "likely broader
than one match arm" and mentioned zero-copy pod handling; measured, it is
not. `Address` is already in scope through the `use quasar_lang::prelude::*`
the scaffold emits, so no import change was needed either.

Before: `codegen --target quasar` on the multisig spec, 13 errors, the first
being "cannot find type `Pubkey` in this scope". After: all three examples
compile, run their generated unit tests and proptests, and type-check their
Kani harness.

## The half that matters

Without a gate this recurs silently, so the #294 artifact gate now runs both
targets (`gate_quasar_example`).

That gate is also how this survived, and the mechanism is worth stating: it
regenerated all three examples with `--target anchor` in a tempdir. The gate
built to catch non-compiling generated code had never compiled the code this
repo actually ships. The snapshot suites did not cover it either — they
capture the default (Anchor) target, so the Quasar output was stable and
wrong in exactly the same way every run.

New gate cases: `{escrow,lending,multisig}_quasar_artifacts_compile_and_run`.
All 7 gate tests pass in 76s.

## A second defect the fix exposed

Regenerating the examples first put `Address` into
`tests/integration_tests.rs`, which would not have compiled. That file is a
HOST-side Parallax test, not a Quasar program: it imports
`parallax_svm::prelude::*`, which exports `Pubkey` and no `Address`, and the
program's own `use quasar_lang::prelude::*` is a private import, so
`use program::state::*` does not carry `Address` across either.

Fixed with an explicit `TypeMapContext::ParallaxHost`, because the context is
about the file's import environment rather than the program's target. It
still matches the Quasar state struct it builds: `solana-pubkey` re-exports
`solana_address::Address as Pubkey`, so the two names denote one type. The
enum variant exists so the next person does not "correct" it back to the
Quasar mapping.

I could not compile-verify that file — see below.

## Not fixed here: the Parallax dependency tree does not build

`codegen --all --target quasar` still fails, for an unrelated reason, and
`parallax_integration_gate` is red on `main` today (exit 101, verified by
running it). The generated manifest already carries compatibility pins for
this, and one is a patch level too loose:

```
wincode = { version = "0.5", features = ["derive"] }   # admits 0.5.5
```

`wincode` resolves to 0.5.5, and `solana-transaction` 4.1.5 and
`solana-sysvar` 4.1.0 fail against it (`the trait bound `Signature:
SchemaRead<..>` is not satisfied`, 18 more). `wincode-0.5.3` is what the
pin was written against. The comment on `parallax_dev_dependencies` says the
pins exist so "a fresh resolve picks the wincode-0.6 line and fails to
compile" — 0.6 was guarded, a 0.5 patch was not.

Likely fix is pinning `=0.5.3` in `parallax_dev_dependencies` and in
`tests/fixtures/parallax-api-gate/stub/Cargo.toml`, needing a full Parallax
build to confirm. Left out of this PR: it is a different defect, and #371's
pin-liveness script checks that the revision EXISTS, not that it builds, so
it stays green through this.

The Quasar gate therefore generates the scaffold plus the three
framework-neutral harnesses rather than using `--all`. Bundling a broken
upstream dependency into this gate would make a Quasar codegen regression
indistinguishable from an upstream break.

## Tests

- Two unit tests pinning the mappings: Quasar → `Address`, Anchor →
  `Pubkey`, standalone → `[u8; 32]`, and ParallaxHost → `Pubkey` with every
  other type identical to Quasar.
- Three new artifact-gate cases (`#[ignore]`, existing CI job picks them up).
- Bundled examples regenerated via `check --regen-drift --write`: 14 files,
  55 lines, every one of them `Pubkey` → `Address` and nothing else.
  `check --regen-drift` is clean.

Full suite green (39 suites, 1636 tests). clippy and rustfmt clean.

## Notes

Stacked on #377 (which is stacked on #373). #377 adds `verify --scaffold`,
whose Quasar examples fail without this fix, so these merge together:
#373 → #377 → this.

`docs/framework-support.md`: the Quasar scaffold row moves to gated, the
Quasar note is rewritten from "does not compile today" to how it survived,
and the Parallax note now records that its gate is red and why.


abishekk92 (commented):
The incremental type-mapping change looks correct: Quasar program code uses Address while the Parallax host scaffold retains Pubkey, and the new generated-artifact gates cover both contexts. No additional blocker found in this diff. It is not ready to merge as stacked, however, because it inherits the red generated-artifacts and runtime-journey jobs from #377; fold #379 into the parent and rebase this PR.

Issue #366, opened 2026-07-27
G23: integration-test scaffold has no Anchor instruction-builder adapter — the lane is Quasar-only


abishekk92:
## Summary

`codegen --integration` emits the Parallax/LiteSVM integration scaffold only
for `--target quasar`. Anchor and Pinocchio get a skip note and no artifact,
so the dominant brownfield target cannot use the lane at all.

Everything on the Parallax side is framework-neutral: world setup, program
loading, execution, `Outcome` checks, account fixtures. The only Quasar-bound
part is the instruction builders, which come from the generated Quasar
`program::client` module.

## Evidence

`crates/qedgen/src/run.rs:2147-2168`:

```
} else if !matches!(target, Target::Quasar) {
    // The execution harness is Parallax-backed, but its instruction
    // builders still come from the generated Quasar program::client
    // module. It does not yet compile inside Anchor or Pinocchio crates
    // — skip with a note instead of writing an artifact that can't build.
    ...
    eprintln!("note: skipping integration-test codegen for {:?} target — \
               the Parallax harness uses Quasar client builders today.", target);
```

`docs/framework-support.md`, integration-tests row: "❌ adapter pending" for
Anchor, Pinocchio, native, and sBPF.

## Proposed fix

Add an Anchor instruction-builder adapter and select the builder by target,
leaving the rest of the scaffold untouched:

- discriminator: `sha256("global:<handler_name>")[..8]`
- args: Borsh-encoded in declaration order
- accounts: the handler's `accounts` block in order, with signer/writable
  flags the scaffold already knows

Codegen already computes exactly this for the Parallax reproducer lane
(`crates/qedgen/src/codegen/parallax_repro.rs`), which builds Anchor-shaped
instructions directly. Factoring that encoder into a shared builder and
calling it from `integration_test` is the smallest change that unblocks the
target — no new dependency and no second copy of the discriminator rule.

Pinocchio needs its own dispatch convention (leading discriminant byte) and
can follow the same seam later.

## Gate

Extend the integration compile gate with an Anchor fixture: generate the
scaffold for an Anchor spec and `cargo check --tests` it against the pinned
Parallax, exactly as the Quasar case does today.

PR #379, opened 2026-07-28
fix(codegen): pin the solana crates that crossed the wincode 0.6 boundary

abishekk92:
`parallax_integration_gate` was red on `main` (exit 101) with no qedgen
change behind it. This turns it green and records the mechanism, because it
will recur.

## What actually broke

Not the pin liveness, and not a version conflict. `litesvm` 0.15 requires
`wincode ^0.5.5`, so the whole graph has to stay on wincode 0.5. Four small
solana crates crossed to `wincode` 0.6 in a MINOR bump:

| crate | last 0.5 | first 0.6 |
|---|---|---|
| `solana-rent` | 4.3.0 | 4.4.0 |
| `solana-signature` | 3.4.1 | 3.5.0 |
| `solana-epoch-schedule` | 3.2.0 | 3.3.0 |
| `solana-fee-calculator` | 3.2.2 | 3.3.0 |

Every parent asks for them with a caret, so a fresh resolve took the
crossing version and put two incompatible `wincode` majors in one graph.

It does not surface as a version error — both wincodes build fine. The
derive on `solana-transaction` targets 0.5's `SchemaRead` while
`solana-signature` implements 0.6's, and rustc reports an unsatisfied trait
bound deep inside a dependency:

```
error[E0277]: the trait bound `Signature: SchemaRead<'_, __WincodeConfig>` is not satisfied
   --> solana-transaction-4.1.5/src/versioned/mod.rs:78:40
note: there are multiple different versions of crate `wincode` in the dependency graph
```

That last note is the only clue, and it is 8 lines into a 28-error wall.

## Fix

Four `=` pins added to `parallax_dev_dependencies`, alongside the nine that
were already there for exactly this reason. Gate goes 101 → 0.

Two hypotheses I tried and discarded first, since the doc comment now
records them: pinning `wincode` itself down to `=0.5.3` is impossible
(litesvm requires `^0.5.5`), and removing the existing `=` pins does not
help (still 28 errors) — they were not the cause.

The doc on `parallax_dev_dependencies` now says this list is not a stable
set but "every crate that has ever crossed the boundary", and gives the
command to find the next one: `cargo tree -i wincode@0.6.0 --depth 1`.

## Two lists collapsed to one

`PARALLAX_DEV_DEP_NAMES` was a hand-kept copy of the keys
`parallax_dev_dependencies` emits — the #363 bug class in miniature, and
this change would have had to touch both. A name in the emitted block but
missing from the list is not detected as subtabled, so upserting a manifest
that declares it as `[dev-dependencies.<name>]` writes a duplicate key and
the user's manifest stops parsing.

Now derived from the emitted block, with a test that the derivation returns
every key and nothing else (no comment, blank, or `[subtable]` line leaking
in as a dependency name).

## A silently disabled test, found on the way

Adding the test above initially consumed the `#[test]` attribute belonging
to `parallax_gate_fixture_pins_the_same_revision`, leaving that function
without one. Clippy's `never used` warning caught it. Restored; both now run.

Worth noting because that test is itself a guard against a silent gate loss
(fixture pinning an older Parallax rev than codegen emits), so it losing its
own attribute would have been the same failure one level up.

## Still broken, not fixed here

`codegen --all --target quasar` produces 10 errors in the generated
`tests/integration_tests.rs` — down from 28, since the other 18 were the
dependency problem. They are a separate codegen defect:

```
error[E0560]: struct `multisig::state::MultisigAccount` has no field named `creator`
   --> tests/integration_tests.rs:66:9
   = note: all struct fields are already assigned
```

The `state_account` helper builds the account struct with a field-shorthand
literal, but Quasar's `#[account(discriminator = N)]` rewrites the struct, so
the literal does not typecheck against the macro's output. `#[allow(dead_code)]`
does not save it — dead code still has to compile.

The gate does not catch this because its fixture is a hand-written stub whose
state struct is not `#[account]`-derived. Filing separately; it belongs with
the integration-lane work in #366.

## Tests

- `dev_dep_names_match_the_emitted_block_exactly`, covering the derivation
  and asserting the four boundary pins are present.
- `parallax_integration_gate` green (was exit 101).
- `generated_artifact_gate` green, 7 cases.
- Bundled examples regenerated: the four pins land in each `Cargo.toml` and
  in the header comment the scaffold emits, which mirrors the dev-dep block.

Full suite green (39 suites, 1637 tests). clippy and rustfmt clean.

## Notes

Stacked on #378 → #377 → #373. Merge in that order.


abishekk92 (commented):
The pin set and the derived dependency-name list look correct, and this PR turns the previously red generated-artifacts and runtime-journey jobs green. Recommendation: move or squash this commit into #377, then rebase #378 and later children. Leaving the fix as a later child means the declared parent merge order includes two red PRs.

PR #380, opened 2026-07-28
fix(codegen): single resolver for emitted error variants; gate the closure

### LINKED ISSUE #363: G21: error-variant declaration and use are two lists — requires/abort names emit undeclared variants and check reports 0 errors

abishekk92:
Closes #363.

Half of this issue was already fixed on the branch by 9922f2e, which the
issue predates. Measured against the current binary, all four user-written
reference sites now report as errors:

| site | status |
|---|---|
| `requires … else X` | ✅ Error (9922f2e) |
| `match` arm `abort X` | ✅ Error (desugars to a synthetic `requires`) |
| effect `else X` | ✅ Error (9922f2e) |
| `checked_{over,under}flow_error` pragma | ✅ Error (9922f2e) |

So the lint gap is closed. This PR does the structural half the issue also
asked for, which is what stops the class recurring.

## One resolver

`emit_errors` built the declared set from `error_codes` plus three synthesis
predicates that lived inside it as locals, so no other emitter could ask what
the enum would contain. Each one guessed, and they guessed in opposite
directions:

- checked arithmetic named `MathOverflow` whether or not the spec declared
  it, and the enum did not carry it: non-compiling program, `check` silent;
- `integration_test::authorization_error` went the other way and deliberately
  UNDER-approximated, refusing to name `InvalidLifecycle` because it could
  not tell whether the enum would have it.

New `codegen_shared::error_variants` owns the answer.
`emit_errors` and `authorization_error` both read it.

## What that fixes, concretely

The bundled multisig spec declares neither `Unauthorized` nor
`InvalidLifecycle`, but has non-init pre-statuses, so `codegen_mir`
synthesizes `InvalidLifecycle` into its enum (`InvalidLifecycle = 7`, in the
committed `errors.rs`). `authorization_error` could not see that and degraded
every forged-signer test to a weak assertion:

```rust
-    // AGENT: the spec declares no authorization error, so this only
-    // asserts that SOME error fired — it also passes on a deserialization
-    // or missing-account failure.
-    assert!(outcome.is_err(), "propose should reject wrong creator");
+    // propose must reject a forged creator with the spec's authorization error.
+    outcome.check(Outcome::error(
+        program::errors::MultisigError::InvalidLifecycle,
+    ));
```

Five assertions across multisig's integration tests were passing for the
wrong reason. `None` is still reachable and still degrades — it just no
longer fires spuriously, and there is now a test for each direction.

## The gate

`tests/error_variant_closure.rs`: for each bundled example at each target,
run codegen and assert every `<Prog>Error::<Variant>` token under `src/`
resolves to a variant the generated `errors.rs` declares. 10 example/target
pairs across Anchor, Quasar, and Pinocchio.

This closes the class rather than the instances — a reference site added
tomorrow is covered without anyone remembering to extend a list. Fast
(codegen only, no cargo), so it runs in the normal suite.

Deliberately NOT written against `emitted_error_variants`, which would be
circular: if the resolver were wrong, emitter and check would agree and the
test would pass while the crate failed to build. It reads the generated
`errors.rs`, which is what `cargo build` sees.

It also carries `the_scan_detects_an_undeclared_reference`, which generates
the live #363 shape and asserts the scan finds it. A gate that cannot fail
reads as coverage and provides none.

## One thing from the issue I did not do

The issue's step 2 says to make the `unknown_error_variant` lint consume the
resolver too, and step 3 recommends switching policy to "synthesize
user-written names + informational lint". Both are superseded by 9922f2e,
which chose the opposite and better policy: report user-written names, and
synthesize only codegen's own defaults.

Feeding the resolver to that lint would actively break it — a misspelled
`else Unathorized` would become a synthesized variant no guard ever raises,
which is a silent behavior change instead of a compile error. The module doc
records this so the next reader does not "finish" the issue by wiring it up.

## Tests

- `error_variant_closure`: 10 example/target pairs + the scan self-test.
- `authorization_error_tracks_the_emitted_enum`, which also asserts the
  emitter really carries what the scaffold names, so the two cannot drift.
- `authorization_error_degrades_when_the_enum_carries_neither`, keeping the
  weak-form path covered.
- Multisig example regenerated; `check --regen-drift` clean.

Full suite green (40 suites, 1649 tests). clippy and rustfmt clean.

## Notes

Stacked on #379 → #378 → #377 → #373. Merge in that order.


abishekk92 (commented):
Blocking review. The shared emitted_error_variants resolver is a good structural fix, but the synthesized-only case is still broken. emit_errors returns immediately when mir.errors.variants is empty before calling the resolver, and emit_lib uses the same declared-only predicate for pub mod errors. A spec with checked arithmetic or synthesized InvalidLifecycle and no declared Error variants therefore references an enum/module that is never emitted. Compute emitted_error_variants first, gate both sites on that resolved set, and add a regression that generates or compiles checked arithmetic with no type Error block.

PR #381, opened 2026-07-28
fix(codegen): mark the placeholder program_id and refuse it in the repro lane

### LINKED ISSUE #368: B16: a spec without program_id generates declare_id! = the System Program address, with no lint
## Summary

When a spec declares no `program_id`, codegen silently falls back to
`11111111111111111111111111111111` — the System Program address — and emits
it as the program's own `declare_id!`. Nothing warns at spec time, the
generated line is not marked as a placeholder, and downstream tooling reads
that value back as the address to send transactions to.

## Repro

Spec with no `program_id` line:

```
$ qedgen check --spec e.qedspec        # no finding about program_id
$ grep -n declare_id programs/src/lib.rs
11:declare_id!("11111111111111111111111111111111");
```

Observed: the generated program declares the System Program's address, and
`check` says nothing.
Expected: a lint, and/or a generated comment marking the value as a
placeholder.

## Root cause

`crates/qedgen/src/codegen/codegen_mir.rs:496-499`:

```rust
let program_id = parsed
    .program_id
    .as_deref()
    .unwrap_or("11111111111111111111111111111111");
```

`crates/qedgen/src/project/init.rs:381` templates the same constant into
newly created specs, so the default path produces it too.

There is no `program_id` lint: `grep -rn program_id crates/qedgen/src/check/lints/`
returns only fixture text.

## Downstream impact

The reproducer lane discovers the deployed program by scanning `src/lib.rs`
for `declare_id!("…")`. With the placeholder in place it resolves the System
Program as the attack target, so a reproducer can report "no bug" for a
reason unrelated to the finding. The repro gate works around this by
rewriting the line before it runs:

`crates/qedgen/tests/parallax_repro_gate.rs:59-74` —
"The scaffold stamps the System Program id as a `declare_id!` placeholder,
which cannot be the program's own address at runtime."

## Proposed fix

1. `missing_program_id` lint (Warning): "no `program_id` declared — the
   generated `declare_id!` falls back to the System Program address. Add
   `program_id \"<your id>\"`, from `anchor keys list` or
   `solana-keygen pubkey target/deploy/<name>-keypair.json`."
2. Keep the fallback (codegen must emit something) but emit it with a
   `// PLACEHOLDER: no program_id in the spec; replace before deploying`
   comment directly above `declare_id!`.
3. Make the reproducer lane refuse a placeholder address with an explicit
   reason instead of aiming transactions at the System Program.

## Gate

A lints test for (1) and a codegen snapshot pinning the placeholder comment
for (2).

abishekk92:
Closes #368.

A spec with no `program_id` gets `11111111111111111111111111111111` — the
System Program's address — stamped as its own `declare_id!`, with nothing
saying so. The trap is that it is a **valid base58 pubkey**, so nothing
downstream rejects it on shape and it reads back as a real address.

## Three fixes, one shared constant

`codegen_shared::PLACEHOLDER_PROGRAM_ID`. Three places have to agree about
this value and they were independent, which is how the repro lane ended up
happily resolving the System Program as an attack target.

1. **Codegen marks it.** The emitted `lib.rs` carries a `// PLACEHOLDER:`
   comment above the line, saying what it is and how to replace it. That
   comment is the only in-band signal a reader or a grep has.
2. **`check` reports it** as P3 `missing_program_id`.
3. **The probe reproducer lane refuses it.** Previously it would resolve the
   System Program as the target, send the attack there, watch it fail for an
   unrelated reason, and report "no bug" — a silent wrong answer.

## Two things the issue did not anticipate

**The lint has to cover a declared placeholder, not just a missing one.**
`project/init.rs` templates that exact value into every new spec, so the
default authoring path produces a spec that *declares* the placeholder.
Checking `program_id.is_some()` would have missed it — and missed three
bundled examples that write it verbatim. Both roads are reported, with
different wording.

**Info, not the Warning the issue proposed.** `SeverityCounts::fails_check`
is `errors > 0 || warnings > 0`, so a Warning makes `qedgen check` exit 1. A
spec has no deployed address for most of its life, so a Warning would fail
`check` on ordinary in-progress work — including the documented quickstart
lane, which asserts a clean check. That gets a rule muted, not obeyed.

The load-bearing protections here are structural (the emitted comment, the
repro-lane refusal); the lint names the situation and carries the fix, which
is what Info is for. I first tried Warning, hit two failing tests, and
checked *why* rather than adjusting them — the exit policy is the real
argument, not the tests.

## A workaround removed at the root

`parallax_repro_gate` rewrote `declare_id!` after generation, with a comment
explaining that the scaffold stamps an address that "cannot be the program's
own at runtime". The fixture spec now declares a real `program_id`, so
codegen emits it and the rewrite is gone — replaced by an assertion, so if
the spec ever loses it the gate fails there instead of the lane silently
reporting "no bug".

Gate re-run with `--ignored`: both reproducers still fire against the real
address (203s, 2 passed).

## Tests

- `missing_program_id_covers_absent_and_placeholder` — absent warns,
  verbatim placeholder warns, a real address is clean.
- `missing_program_id_emits_a_marked_placeholder` and
  `declared_program_id_gets_no_placeholder_comment` — the comment appears
  exactly when the fallback is used.
- `placeholder_program_id_is_refused` — the repro lane fails closed with a
  reason that says both why and how to fix it.
- Two snapshots refreshed; the diff is the 5-line comment and nothing else.

No existing test was weakened. I briefly added an exemption to
`test_complete_spec_clean` while the lint was a Warning and reverted it once
Info made it unnecessary.

Full suite green (40 suites, 1653 tests). clippy and rustfmt clean.

## Bundled examples now carry real ids

Every example declares a `program_id` generated with `solana-keygen`, one
per example, with the imported `cross-program-vault-admin` getting its own.
They demonstrate the practice rather than the placeholder. Keypairs were
discarded; only the pubkeys are checked in.

That change is wider than it looks, and the reason is worth recording:
`program_id` is a top-level spec item, so it feeds `spec_context_digest`
and therefore **every handler's `spec_hash`**. Adding one re-stamps every
`#[qed(verified, …)]` in that program. Concretely: 29 stamps across escrow,
lending, multisig, and cross-program-vault, plus two expectations in
`crates/qedgen-hash-core/tests/stamp_crosscheck.rs`.

That digest is documented as deliberately conservative ("a top-level change
NO handler references still invalidates every hash"), so this is by design,
not a defect. `docs/limitations.md` now warns maintainers that a
`program_id` edit is a re-stamping change.

Re-stamped via `spec_hash_for_handler` directly rather than by hand, and the
pass is idempotent (a second run is a no-op).

One thing worth a look: `qedgen check --drift` reported `8 verified, 0
drifted` on multisig while its `spec_hash` values were genuinely stale — the
crosscheck test is what caught it. The `spec = "../multisig.qedspec"` path in
the stamp resolves relative to the stamped file, which does not land on the
spec, so the spec leg appears to be skipped rather than failed. I have not
filed that yet; it is adjacent to this issue but a separate defect.

## Notes

Stacked on #380 → #379 → #378 → #377 → #373. Merge in that order.


abishekk92 (commented):
The placeholder constant, in-band marker, lint, and fail-closed reproducer check are coherent, and CI is green. I found no new blocking issue in this incremental diff. Merge only after the parent stack blockers are repaired, especially the synthesized-only error emission defect in #380; the broad program_id restamp churn should then be regenerated once on the final base.

PR #384, opened 2026-07-28
fix(check): detect Pinocchio in regen-drift so its program crate is compared

### LINKED ISSUE #367: B15: target_from_text cannot return Target::Pinocchio, so regen-drift silently skips Pinocchio program crates
## Summary

The framework detector behind `check --regen-drift` recognizes `anchor-lang`
and `quasar-lang` only. `Target` is a closed three-variant enum, but
`target_from_text` can never return `Target::Pinocchio` — a Pinocchio project
resolves to `None`, and the caller treats `None` as "skip this crate".

Effect: regen-drift never regenerates or compares a Pinocchio program's
`src/*.rs`. Drift between the spec and a generated Pinocchio scaffold is
invisible, and the gate reports success.

Second effect: a newer consumer of the same detector had to fail closed
against `detected: None` rather than get a real answer, because "unknown"
cannot be distinguished from "Pinocchio".

## Evidence

`crates/qedgen/src/verify/regen_drift.rs:355-363`:

```rust
fn target_from_text(body: &str) -> Option<Target> {
    if body.contains("anchor-lang") || body.contains("anchor_lang::prelude") {
        Some(Target::Anchor)
    } else if body.contains("quasar-lang") || body.contains("quasar_lang::prelude") {
        Some(Target::Quasar)
    } else {
        None
    }
}
```

`regen_drift.rs:296-310` — the skip:

```rust
let Some(target) = detect_program_target(&out_root)? else {
    continue;                       // Pinocchio crate: never regenerated
};
outputs.push((rel, target));
```

Pinocchio detection already exists twice elsewhere in the codebase:

- `crates/qedgen/src/probe/mod.rs:1382` — `has_pinocchio_dep(cargo_toml)`
- `crates/qedgen/src/adapt/mod.rs:57` — `detect_adapter` covers
  Anchor / Pinocchio / native

## Repro

1. `qedgen codegen --spec x.qedspec --target pinocchio`
2. hand-edit a line in the generated `programs/src/guards.rs`
3. `qedgen check --regen-drift`

Observed: no drift reported.
Expected: the edited generated file is reported as drifted.

## Proposed fix

One shared `detect_framework(root) -> Option<Target>` in a single module,
consumed by `regen_drift`, the reproducer lane, and `adapt`. Give it a
Pinocchio arm keyed on the `pinocchio` Cargo dependency (reuse
`has_pinocchio_dep`) or on `pinocchio::` / `pinocchio_pubkey::declare_id!` in
`src/lib.rs`. Three detectors for one question is the underlying problem; the
missing arm is the symptom.

## Gate

A `regen_drift` test that a generated Pinocchio project round-trips with no
drift, and that an edited generated file in one IS reported as drifted.

abishekk92:
Closes #367.

`Target` is a closed three-variant enum, but `target_from_text` had two arms
and an `else None`, so a Pinocchio crate read as "no target" — and
`program_outputs` treats `None` as "skip this crate".

## The symptom is louder than the issue says

The issue describes a silent skip: *"Drift ... is invisible, and the gate
reports success."* Measured, it is worse than that. With **zero** drift, a
Pinocchio project reports five findings:

```
$ qedgen check --spec pino/ping.qedspec --regen-drift --examples-root exroot
Example codegen drift detected:
  pino: programs/Cargo.toml (missing generated counterpart)
  pino: programs/src/errors.rs (missing generated counterpart)
  pino: programs/src/guards.rs (missing generated counterpart)
  pino: programs/src/instructions/mod.rs (missing generated counterpart)
  pino: programs/src/state.rs (missing generated counterpart)
```

Nothing regenerated into the temp tree, so every committed file has no
counterpart to compare against. Planting real drift in `guards.rs` produces
the same output. So drift is genuinely invisible — not because the gate
passes, but because a clean project and a drifted one are indistinguishable,
both buried under the same false findings.

The issue's repro also assumes a Pinocchio example exists under an examples
root; none does, which is why this never showed up in CI.

## After

```
$ qedgen check ... --regen-drift --examples-root exroot
Example codegen drift check clean — 1 example(s) checked.

$ echo "// DELIBERATE DRIFT" >> pino/programs/src/state.rs
$ qedgen check ... --regen-drift --examples-root exroot
Example codegen drift detected:
  pino: programs/src/state.rs (changed)
```

## Fix

One arm per `Target` variant, iterating the variants so the match is
exhaustive and **the `_ => None` fallback is gone**. A fourth variant now
fails to compile here rather than quietly becoming "undetectable", which is
the shape that produced this bug (the same closed-enum discipline CLAUDE.md
describes for `Stmt` and `SeverityCounts`).

The Pinocchio predicate reuses `probe::has_pinocchio_dep` rather than adding
a third copy. That function is line-oriented, so it matches
`pinocchio = "0.8"` / `pinocchio.workspace` / `pinocchio-token` in a manifest
without firing on a passing mention; `pinocchio::` and `pinocchio_pubkey::`
cover `src/lib.rs`.

## What I did not do

The issue proposes collapsing the three framework detectors into one shared
`detect_framework`. I did not, because they answer different questions and
`docs/framework-support.md` opens by saying so: "Five *distinct* framework
notions exist — they are not one enum." `regen_drift` wants the greenfield
`Target`; `adapt::detect_adapter` wants `ProgramFramework` (anchor /
pinocchio / **native**, no quasar). Merging them would force one enum onto
two different domains. Sharing the Pinocchio predicate gets the
de-duplication that was actually available.

## Second effect, now accurate

`probe_repro` fails closed on `detected != Some(Target::Anchor)` and its
comment explained it had to treat "unknown" as "not Anchor" because Pinocchio
read as `None`. It now reports `Some(Pinocchio)` — same refusal, but the
reason names the framework instead of saying nothing was recognised. Comment
updated to match.

## Tests

- `detects_every_target_from_manifest_and_lib` — all three targets from both
  a manifest and a `src/lib.rs`, in the exact shapes codegen emits, plus an
  unrelated crate still reading `None` (which is what lets `program_outputs`
  skip non-generated crates).
- `pinocchio_arm_does_not_fire_on_a_mention` — a manifest comment, a
  `not-pinocchio` dependency, and an Anchor `lib.rs` whose prose mentions
  `pinocchio::`. A false positive here would be worse than the miss it
  replaces: it would regenerate an Anchor crate as Pinocchio and report every
  file as drifted.

Worth noting: my first draft of that second test asserted
`pinocchio-like-thing` should not match. It does, and correctly —
`has_pinocchio_dep` counts `pinocchio-*` siblings on purpose since they
require the root pinocchio surface. I fixed the test rather than the shared
function.

Full suite green (40 suites, 1655 tests). clippy and rustfmt clean.

## Notes

Stacked on #381 → #380 → #379 → #378 → #377 → #373. Merge in that order.

### DIFF
--- a/crates/qedgen/src/probe/mod.rs
+++ b/crates/qedgen/src/probe/mod.rs
@@ -1379,7 +1379,12 @@ fn detect_runtime(root: &Path) -> Runtime {
 /// Pinocchio dep check. Matches `pinocchio = ...`, `pinocchio.workspace =
 /// true`, or `pinocchio-token`/`-system` siblings (siblings require the
 /// root pinocchio surface).
-fn has_pinocchio_dep(cargo_toml: &str) -> bool {
+///
+/// Line-oriented rather than a bare `contains("pinocchio")` so a mention in
+/// a comment or inside a longer crate name does not count. Shared with
+/// `verify::regen_drift` (#367) — a second copy there would be a third
+/// framework detector answering the same question a different way.
+pub(crate) fn has_pinocchio_dep(cargo_toml: &str) -> bool {
     for line in cargo_toml.lines() {
         let t = line.trim();
         if t.starts_with('#') {
--- a/crates/qedgen/src/probe/probe_repro.rs
+++ b/crates/qedgen/src/probe/probe_repro.rs
@@ -228,11 +228,15 @@ fn discover_deployed_program(project_root: &Path) -> Result<DeployedProgram, Con
             .flatten()
     });
     // FAIL CLOSED: require positive Anchor detection rather than refusing
-    // only the frameworks we happen to recognise. `target_from_text` knows
-    // Anchor and Quasar, so a Pinocchio project reads as `None` — treating
-    // "unknown" as "probably fine" would emit an Anchor-shaped harness
-    // against it and report "no bug" from a transaction that never reached
+    // only the frameworks we happen to recognise. Treating "unknown" as
+    // "probably fine" would emit an Anchor-shaped harness against another
+    // framework and report "no bug" from a transaction that never reached
     // the handler.
+    //
+    // The detector now covers all three targets (#367), so a Pinocchio
+    // project reports `Some(Pinocchio)` here instead of the `None` it used
+    // to — same refusal, but the reason names the framework rather than
+    // saying nothing was recognised.
     if detected != Some(crate::Target::Anchor) {
         return Err(ConstructFailure::BuildError(format!(
             "Parallax reproducers are Anchor-only today (detected: {detected:?}). The \
--- a/crates/qedgen/src/verify/regen_drift.rs
+++ b/crates/qedgen/src/verify/regen_drift.rs
@@ -352,14 +352,44 @@ pub(crate) fn detect_program_target(out_root: &Path) -> Result<Option<Target>> {
     Ok(lib_target.or(cargo_target))
 }
 
+/// Which greenfield target generated this `Cargo.toml` or `src/lib.rs`?
+///
+/// One arm per [`Target`] variant. `Target` is a closed three-variant enum,
+/// but this had only two arms and an `else None` (#367), so a Pinocchio
+/// crate read as "no target" — and `program_outputs` treats `None` as "skip
+/// this crate". The effect was not a silent pass: nothing regenerated into
+/// the temp tree, so every committed Pinocchio file reported as `missing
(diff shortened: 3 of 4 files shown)


abishekk92 (commented):
The exhaustive Target mapping fixes the Pinocchio regen-drift blind spot, reuses the shared dependency detector, and has focused positive and negative coverage. CI is green and I found no new blocking issue in this incremental diff. Merge after the repaired parent chain.

PR #386, opened 2026-07-28
fix(proptest): sequence restores the pre-state when an op is rejected

### LINKED ISSUE #385: proptest: rejected transitions leave partial state, which the sequence harness then carries forward
## Summary

Generated transition functions mutate `State` field-by-field and `return false` mid-way when a later effect fails its checked arithmetic. The already-written fields stay mutated. In `state_machine_sequence` the rejected op's partial state then carries into the next op, so the harness explores states no legitimate transition can produce.

## Repro (generic)

```
spec Atomic

state { a : U64, b : U64 }

type Error | Bad

handler bump (x : U64) {
  requires x > 0 else Bad
  effect {
    a += x
    b += x
  }
}
```

Emitted transition:

```rust
fn bump(s: &mut State, x: u64) -> bool {
    if !(x > 0) { return false; }
    match s.a.checked_add(x) { Some(__v) => s.a = __v, None => return false }
    match s.b.checked_add(x) { Some(__v) => s.b = __v, None => return false }
    //                                      ^^^^^^^^^^^^ `s.a` is already mutated here
    true
}
```

With `b` near `u64::MAX`, `a += x` commits and `b += x` rejects — `bump` returns `false` having half-applied its effect.

## Why it matters

The sequence loop keeps the mutated state on a rejected op:

```rust
let pre = s;
if apply_op(&mut s, op) {   // false -> falls through, `s` stays partially mutated, `pre` discarded
    ...
}
```

So a later op runs against a state the model never legitimately reached. That can produce a false counterexample (a property violated by a half-applied effect) or mask a real one.

Scope: this is only observable in `state_machine_sequence` — it is the one harness that reuses `s` across ops. The preservation, guard-rejection, and overflow harnesses either assert nothing on a `false` return or assert only the return value, so a partial write there is inert.

There is no rollback in the emitter today: the only snapshot (`parallel_snapshot_fields` in `rust_codegen_util/emit.rs`) implements parallel effect semantics (RHS reads see pre-state), not failure rollback.

## Suggested fix

Two options, in increasing blast radius:

1. **Sequence-local rollback (minimal).** Restore the pre-state when the op is rejected — semantically exact, since a rejected transition must not mutate. `pre` is already captured immediately before `apply_op` (added in #375), so this is roughly `else { s = pre; }` in `emit_sequence_test_for` plus `emit_product_sequence_test`, with the `let pre = s;` emission made unconditional.
2. **Atomic transitions (thorough).** Compute every new field value into locals and commit them only after all effects succeed. Correct for any future consumer that reuses state after a rejection, but a larger change to the shared emitter and needs Kani-lane parity.

Option 1 looks like the right size for the actual exposure.

## Related, lower priority

`max_global_rejects` is a hard-coded `65536` in the preservation/guard/overflow harnesses. #376 removed most of the rejection pressure for conservation shapes, but deep runs (tens of thousands of cases) on shapes the repair does not recognize can still exhaust it. Making it scale with the case count — or overridable via env like `PROPTEST_CASES` — would avoid an abort that reads as a failure.

saicharanpogul:
Closes #385.

## The bug

Generated transition fns assign field-by-field and `return false` from the middle when a later effect fails its checked arithmetic — the fields already written stay written:

```rust
fn bump(s: &mut State, x: u64) -> bool {
    if !(x > 0) { return false; }
    match s.a.checked_add(x) { Some(__v) => s.a = __v, None => return false }
    match s.b.checked_add(x) { Some(__v) => s.b = __v, None => return false }
    //                                      ^^^^^^^^^^^^ `s.a` is already mutated here
    true
}
```

`state_machine_sequence` is the one harness that reuses `s` across ops, and it discarded `pre` on a rejected op:

```rust
let pre = s;
if apply_op(&mut s, op) {   // false -> falls through, `s` stays half-applied
    ...
}
```

So every later op ran against a state no real transition can produce. That can fabricate a counterexample *and* mask a genuine one — the corrupted pre-state silently fails the `pre -> post` guard, so the property quietly stops being checked for the rest of the run.

## The fix

Snapshot the pre-state unconditionally and restore it when `apply_op` returns false, in both `emit_sequence_test_for` and `emit_product_sequence_test`:

```rust
let pre = s;
if apply_op(&mut s, op) {
    …
} else {
    s = pre; // rejected op must not half-apply
}
```

The snapshot already existed for the `preserved_by` guard (#375); it's now emitted regardless of whether the section checks any property, since the rollback needs it either way.

## Scope: why not full atomicity

This deliberately does **not** make the transition fns themselves atomic. The partial write is only observable where state is reused across calls, and the sequence is the only such harness:

| harness | on a `false` return | exposure |
|---|---|---|
| preservation | asserts nothing | inert |
| overflow | asserts nothing | inert |
| guard rejection | asserts the return value only | inert |
| **sequence** | **reuses `s` for the next op** | **fixed here** |

Full atomicity (compute every new value into locals, commit at the end) would touch the emitter shared with the Kani lane for no additional coverage. It stays documented in #385 as the heavier alternative should a future consumer reuse state after a rejection.

## Validation

- Regression test `sequence_restores_pre_state_when_an_op_is_rejected` pins both the half-apply shape it guards against and the emitted rollback.
- Lending / multisig / product-ghost snapshots and the bundled example harnesses pick up the `else { s = pre; }` branch — the only diff.
- The regenerated multisig example proptest still passes all 23 tests at 2000 cases.
- All gates green: `cargo test`, `fmt --check`, `clippy --all-targets -D warnings` (on 1.97, matching CI), `check --regen-drift`, `check-readme-drift.sh`, `release-gate.sh`, `codegen_smoke --ignored`.

### DIFF
--- a/crates/qedgen/src/codegen/proptest_gen_mir.rs
+++ b/crates/qedgen/src/codegen/proptest_gen_mir.rs
@@ -2282,11 +2282,10 @@ fn emit_sequence_test_for(
         })
         .collect();
 
-    // Snapshot the pre-state (State is `Copy`) before the transition so the
-    // check below can read the property at `pre`.
-    if !seq_props.is_empty() {
-        out.push_str("            let pre = s;\n");
-    }
+    // Snapshot the pre-state (State is `Copy`) before the transition: the
+    // property checks below read it, and a rejected op restores from it.
+    // Unconditional — the rollback needs it even with no checked property.
+    out.push_str("            let pre = s;\n");
     out.push_str("            if apply_op(&mut s, op) {\n");
 
     if has_lifecycle {
@@ -2337,6 +2336,13 @@ fn emit_sequence_test_for(
         out.push_str("                }\n");
     }
 
+    // A rejected transition must not mutate the state. The generated
+    // transition fns assign field-by-field and `return false` from the
+    // middle when a later effect fails its checked arithmetic, so the
+    // earlier fields stay written. Restoring the snapshot keeps the
+    // sequence exploring only states a real transition can produce.
+    out.push_str("            } else {\n");
+    out.push_str("                s = pre; // rejected op must not half-apply\n");
     out.push_str("            }\n");
     out.push_str("        }\n");
     out.push_str("    }\n");
@@ -3073,10 +3079,10 @@ fn emit_product_sequence_test(
         })
         .collect();
     out.push_str("            for (i, op) in ops.iter().enumerate() {\n");
-    // Snapshot the pre-state (ProductState is `Copy`) for the `pre -> post` check.
-    if !seq_ghost_props.is_empty() {
-        out.push_str("                let pre = s;\n");
-    }
+    // Snapshot the pre-state (ProductState is `Copy`): the `pre -> post`
+    // checks read it, and a rejected op restores from it. Unconditional —
+    // the rollback needs it even with no checked ghost property.
+    out.push_str("                let pre = s;\n");
     out.push_str("                if apply_op(&mut s, op) {\n");
     out.push_str("                    if !initialized {\n");
     out.push_str("                        initialized = true;\n");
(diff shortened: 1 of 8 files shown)


abishekk92 (approved):
Reviewed against current main. The rollback is correctly applied only on rejected sequence ops in both single-account and product-state harnesses, preserving successful-transition property checks and initialization behavior. Generated snapshots and CI cover both paths; no blockers.

PR #391, opened 2026-07-29
fix(probe): size reproducer fixture bytes from the declared type, not a 32-byte guess

### LINKED ISSUE #389: B20: parallax_repro state_bytes gives every non-U* field 32 bytes, so a wide-state reproducer fails deserialization and reports no bug
## Summary

The Parallax reproducer lane builds the pre-state account with
`zero_field_expr`, which knows only `U8`..`U128` and `Bool`. Every other DSL
type falls through to a 32-byte zero blob.

For `Pubkey` that is correct, and it is the only reason the lane works
today. For a signed integer, `Bytes64`, a `Map[N] T` array, or a record, the
width is wrong, so the account is the wrong length. Anchor rejects it during
deserialization, before the guard under test runs, and the lane reports "no
bug".

That is a false negative in an audit lane: a real finding is silently
downgraded to unconfirmed. The emitter's own doc comment names this exact
failure as the reason the helper exists:

> Without it the program aborts with `AccountNotInitialized` before reaching
> the guard under test, and the reproducer would report "no bug" for
> entirely the wrong reason.

## Evidence

`crates/qedgen/src/codegen/parallax_repro.rs:246-254`:

```rust
fn zero_field_expr(dsl_ty: &str) -> String {
    match crate::codegen::repro_gen::rust_int_type(dsl_ty) {
        Some("u8") => "data.push(0);".to_string(),
        Some(rust_ty) => format!("data.extend_from_slice(&0{rust_ty}.to_le_bytes());"),
        // Pubkey and anything else addressable is 32 bytes.
        None if dsl_ty.trim() == "Bool" => "data.push(0);".to_string(),
        None => "data.extend_from_slice(&[0u8; 32]);".to_string(),
    }
}
```

`rust_int_type` (`repro_gen.rs:34-43`) maps `U8`/`U16`/`U32`/`U64`/`U128`
and returns `None` for everything else, including every SIGNED integer.

| state field type | bytes emitted | bytes the program decodes |
|---|---|---|
| `Pubkey` | 32 | 32 ✅ |
| `U64` | 8 | 8 ✅ |
| `I64` | 32 | 8 ❌ |
| `I8` | 32 | 1 ❌ |
| `Bytes64` | 32 | 64 ❌ |
| `Map[32] Pubkey` | 32 | 1024 ❌ |
| `Map[32] U8` | 32 | 32 (by luck) |

A multisig-shaped state (`members : Map[32] Pubkey`, `voted : Map[32] U8`)
comes out 992 bytes short.

## Why it survived

The repro gate fixture
(`crates/qedgen/tests/fixtures/parallax-repro-gate/vulnerable.qedspec`)
declares exactly two state fields, `owner : Pubkey` and `total : U64` —
the two types the emitter gets right. The unit tests assert those same two
shapes (`parallax_repro.rs:632`, `:635`). No test or fixture exercises a
signed, wide, or array field, so the whole class is invisible.

`arg_zero_expr` (`parallax_repro.rs:445-446`) has the same `None => [0u8;
32]` fallback for instruction arguments and is worth checking in the same
pass.

## Proposed fix

`codegen_shared::emit_pod_bytes_append` (added in #383) already does
width-correct, recursive, alignment-1 byte emission for exactly this type
universe, and refuses types with no fixed layout instead of guessing. Factor
the width logic out of it so the repro lane can emit zeros at the right
widths through the same function.

One caveat to check rather than assume: `emit_pod_bytes_append` encodes the
QUASAR Pod layout, and this lane is Anchor/Borsh. The two agree on fixed
scalars and fixed arrays (little-endian, concatenated, no padding) and on
`bool` as one byte, which covers every type listed above. They diverge on
`Vec` and `Option`, which both sides should refuse anyway.

Refusing is the important half. A type this emitter cannot size should be an
error at generation time, not a silent 32-byte guess that turns into a false
"no bug" later.

## Gate

Widen the repro fixture spec with a signed field and a `Map[N] Pubkey`
field, then assert the emitted `state_bytes` length equals the account size
Anchor computes for that struct. A length assertion catches the whole class;
per-type string assertions only catch the types someone thought to list.

abishekk92:
Closes #389.

The Parallax reproducer installs a pre-state account before driving its
attack. `zero_field_expr` sized that account by asking `rust_int_type`,
which knows `U8`..`U128` and returns `None` for everything else, and gave
every `None` a flat 32 bytes.

That is correct for `Pubkey`, which is the only reason the lane worked. It
is wrong for every signed integer, `Bytes64`, and every `Map[N] T`.

| state field | bytes emitted | bytes decoded |
|---|---|---|
| `Pubkey` | 32 | 32 ✅ |
| `U64` | 8 | 8 ✅ |
| `I64` | 32 | 8 ❌ |
| `Bytes64` | 32 | 64 ❌ |
| `Map[4] Pubkey` | 32 | 128 ❌ |

A wrong-length account fails Anchor deserialization before the guard under
test runs, so the attack is recorded as refused and the finding is dropped.
That is a false negative in the one lane whose entire job is evidence, and
it is the exact failure the emitter's own doc comment says it exists to
prevent.

## Width comes from the declared type

`codegen_shared::fixed_byte_width` derives the width from the DSL type:
scalars, `Bool`, addresses, `Bytes32`/`Bytes64`, `Map[N] T` recursively,
records as the sum of their fields, and aliases through their RHS.

Both layouts this repo emits agree on all of it. Anchor/Borsh and
Quasar/Pod both store fixed scalars little-endian with no padding, `bool` in
one byte, an address in 32, and `[T; N]` as N elements back to back. They
diverge only on the variable-length types, which the function refuses for
both. `Fin[N]` is deliberately absent: Quasar packs it as `PodU32` while
Anchor maps it to `usize`, so there is no single answer to give.

Refusing is half the fix. `generate` is now fallible, and a spec carrying a
type the emitter cannot size produces no reproducer and a stated reason
instead of one built on a guess. A reproducer that cannot be built faithfully
should not exist.

`param_witness` had the same `None => [0u8; 32]` fallback for instruction
arguments, so a `Bool` parameter sent 32 bytes where the program reads 1.
Fixed the same way.

The emitted form now states the width literally
(`data.extend_from_slice(&[0u8; 8]);` rather than `&0u64.to_le_bytes()`), so
a wrong width is visible in the generated file instead of hidden behind a
type name.

## The fixture is the gate

`vulnerable.qedspec` declared `owner : Pubkey` and `total : U64` — the two
types the old code sized correctly. That is why this survived. It now also
carries `members : Map[4] Pubkey` and `delta : I64`.

Verified end to end, both directions, with `cargo build-sbf` + LiteSVM:

- with the fix, `parallax_repro_gate` passes: the reproducers still confirm
  the genuinely unguarded handlers and still drop the guarded one.
- with `zero_field_expr` temporarily reverted to the old behaviour and the
  widened fixture in place, the gate fails with

  ```
  set_fee/missing_signer must be CONFIRMED — the program accepts the attack.
  confirmed: []
  ```

  which is the false negative, reproduced.

Unit coverage: the width table including every signed type, `Map` bounds
(nested), records and aliases, the refusal set, plus a `state_bytes` test
asserting the TOTAL emitted length against the account layout. A total
catches the whole class; per-type string assertions only catch the types
someone thought to list.

Full `cargo test` green (40 targets), clippy and rustfmt clean.

### DIFF
--- a/crates/qedgen/src/codegen/codegen_shared/tests.rs
+++ b/crates/qedgen/src/codegen/codegen_shared/tests.rs
@@ -3285,5 +3285,78 @@ fn flat_state_fields_carry_the_synthesized_tail() {
     // Both synthesized fields are single bytes in the struct.
     for (name, ty) in fields.iter().filter(|(n, _)| n != "total") {
         assert_eq!(ty, "U8", "{name} must map to a single byte");
+
+// ── Fixed byte widths (#389) ─────────────────────────────────────────
+
+/// The widths a fixture emitter has to get right. The old repro-lane
+/// emitter knew `U8`..`U128` and gave everything else 32 bytes, which is
+/// correct for `Pubkey` and wrong for every row below it.
+#[test]
+fn fixed_byte_width_covers_the_scalar_universe() {
+    let spec = empty_spec();
+    let width = |ty: &str| fixed_byte_width(ty, &spec).unwrap();
+
+    for (ty, bytes) in [
+        ("U8", 1),
+        ("U16", 2),
+        ("U32", 4),
+        ("U64", 8),
+        ("U128", 16),
+        ("Bool", 1),
+        ("Pubkey", 32),
+        ("Bytes32", 32),
+    ] {
+        assert_eq!(width(ty), bytes, "{ty}");
+    }
+
+    // Signed integers were the silent half: `rust_int_type` returns `None`
+    // for all of them, so every one used to be sized at 32 bytes.
+    for (ty, bytes) in [("I8", 1), ("I16", 2), ("I32", 4), ("I64", 8), ("I128", 16)] {
+        assert_eq!(width(ty), bytes, "{ty}");
+    }
+
+    assert_eq!(width("Bytes64"), 64);
+}
+
+/// `Map[N] T` is N elements, not one pointer-sized anything.
+#[test]
+fn fixed_byte_width_multiplies_map_bounds() {
+    let spec = empty_spec();
+    assert_eq!(fixed_byte_width("Map[32] Pubkey", &spec).unwrap(), 1024);
+    assert_eq!(fixed_byte_width("Map[3] U8", &spec).unwrap(), 3);
+    // Nested, because a bound is a count of whatever follows it.
+    assert_eq!(fixed_byte_width("Map[2] Map[4] U64", &spec).unwrap(), 64);
+}
+
+/// A record is the sum of its fields; an alias resolves to its RHS.
+#[test]
+fn fixed_byte_width_walks_records_and_aliases() {
+    let mut spec = empty_spec();
+    spec.records.push(crate::check::ParsedRecordType {
+        name: "Entry".to_string(),
+        fields: vec![
+            ("who".to_string(), "Pubkey".to_string()),
+            ("amount".to_string(), "U64".to_string()),
+        ],
+    });
+    spec.type_aliases
+        .push(("Amount".to_string(), "U64".to_string()));
+
+    assert_eq!(fixed_byte_width("Entry", &spec).unwrap(), 40);
+    assert_eq!(fixed_byte_width("Amount", &spec).unwrap(), 8);
+    assert_eq!(fixed_byte_width("Map[2] Entry", &spec).unwrap(), 80);
+}
+
+/// Variable-length types have no fixed width, and `Fin[N]` has one only
+/// once a target is chosen. Both must refuse rather than return a number
+/// that reads as authoritative.
+#[test]
+fn fixed_byte_width_refuses_what_it_cannot_size() {
+    let spec = empty_spec();
+    for ty in ["Vec U64", "Option Pubkey", "Fin[8]"] {
+        assert!(
+            fixed_byte_width(ty, &spec).is_err(),
+            "{ty} must not produce a width"
+        );
     }
 }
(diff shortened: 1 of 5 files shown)


abishekk92:
Addressed both review points, plus a third the first one uncovered.

**1. Cyclic type expansion (confirmed, and `check` is not the backstop).**

The review assumed the spec validator guards this. It does not. `lints/known_types.rs:126` resolves aliases with its own cycle guard, and having stopped early it validates the name it stopped on — which for a cycle is a declared alias, so `type A = B` with `type B = A` lints clean and reaches the width function. A self-referential record (`Node { next : Node }`) does the same.

`fixed_byte_width` now threads an `expanding` set through alias and record expansion and returns "type `A` expands into itself" instead of exhausting the stack.

**2. The same hole existed in `emit_pod_bytes_append`** (shipped in #388, so it was not in this diff). It walks aliases the same way with no guard. Fixed here with its own guard rather than delegating to `fixed_byte_width`: the two accept different type sets, since the value emitter supports `Fin[N]` and the width function deliberately refuses it, so borrowing the width function as a gate would have rejected `type Idx = Fin[8]` in a Quasar state. The test pins both halves.

**3. Byte-count helper scope.** `emitted_byte_count` now reads only the `state_bytes` body via `state_bytes_body`.

Rather than just narrowing it, I made the fragility a covered case: the test fixture handler now takes a `flag : Bool` parameter, whose witness emits `data.extend_from_slice(&[0u8; 1]);` into instruction data. That is exactly the shape the review predicted would fold into the total. Verified both ways — with the scoping removed the test fails `left: 247, right: 246`, and passes with it. The guard is load-bearing now rather than hypothetical.

Full `cargo test` green (40 targets), clippy and rustfmt clean.

Issue #383, opened 2026-07-28
B19: the Parallax integration scaffold builds a struct literal Quasar's #[account] view type does not accept


abishekk92:
## Summary

The Parallax integration scaffold's `state_account` helper builds the
program's state account with a plain struct literal:

```rust
let state = MultisigAccount { creator, threshold, member_count, /* … */ bump };
```

Quasar's `#[account]` macro does not leave a struct with those fields, so
the generated `tests/integration_tests.rs` does not compile.

## Repro

```
$ qedgen codegen --spec multisig.qedspec --target quasar --all --output-dir programs
$ cd programs && cargo check --tests
error[E0560]: struct `multisig::state::MultisigAccount` has no field named `creator`
  --> tests/integration_tests.rs:66:9
   |
66 |         creator,
   |         ^^^^^^^ `multisig::state::MultisigAccount` does not have this field
   |
   = note: all struct fields are already assigned
```

10 errors, all in `tests/integration_tests.rs`, one per state field plus the
`SchemaWrite` bound that follows.

`src/state.rs` does declare them, which is what makes the error confusing:

```rust
#[account(discriminator = 1)]
pub struct MultisigAccount {
    pub creator: Address,
    pub threshold: u8,
    // …
}
```

## Root cause

`#[account]` replaces the annotated struct with a **view type** and moves
the declared fields into a hidden zero-copy companion.
`quasar-derive-0.0.0/src/account/fixed.rs:109-123`:

```rust
#[doc(hidden)]
pub mod #zc_mod {
    #[repr(C)]
    #[derive(Copy, Clone)]
    pub struct #zc_name { #(#zc_fields,)* }   // the declared fields, Pod-typed
}
```

The user-facing `#name` keeps a single `__view` field — hence "all struct
fields are already assigned" — and the macro provides the write path
(`fixed.rs:127`):

```rust
impl #name {
    pub fn set_inner(&mut self, #(#field_names: #field_types),*) { … }
}
```

So the fields exist, just not as constructible struct fields. The scaffold
is written against the pre-macro shape.

`crates/qedgen/src/codegen/integration_test.rs` (the `state_account`
emitter) builds the literal and then hands it to `wincode::serialize(&state)`
to produce account bytes.

## Why it survived

`parallax_integration_gate` compiles the scaffold against a hand-written
stub crate whose state struct is a plain struct, not `#[account]`-derived,
so the shape the real macro produces is never exercised. Same class as #372:
[...]

Issue #279, opened 2026-07-19
codegen output-dir defaults resolve against CWD, not the spec — stray artifacts when invoked from outside the project


abishekk92:
## Summary

Every codegen output-dir default is a literal relative path (`cli.rs`: `default_value = "./formal_verification"`, `./programs`, …) resolved against the *current working directory*, not the `--spec` location. Running `qedgen codegen --proptest --spec examples/rust/escrow/` from the repo root silently writes a fresh `./programs/` tree (9 files) at the repo root instead of updating the example — the command 'succeeds' and the real target stays stale.

Surfaced by the #272 workaround-prose sweep: this defect has been *worked around in writing twice* without being filed —

- `references/release-history.md` (v2.27.1): the lake-build workflow had to `cd` into each example before `qedgen codegen --lean --spec .` because `--lean-output`'s cwd-relative default made the lakefile inject step short-circuit silently from repo root;
- 2026-07-19 session: the stray repo-root `programs/` above, while refreshing a drifted example artifact (recovered via `check --regen-drift --write`, which resolves paths correctly).

## Expected

Relative output-dir defaults (and arguably relative user-supplied values) should resolve against the spec's project root — the directory containing the `.qedspec` / `qed.toml` — matching what `check --regen-drift` already does. Invoking any codegen from anywhere with `--spec <path>` should write into that project.

## Proposed fix

Resolve default output dirs against the spec dir at the dispatch layer (one seam, all subcommands), keeping absolute and explicitly-passed paths untouched. Gate: a test invoking `codegen --proptest --spec <fixture-dir>` from a *different* cwd and asserting artifacts land under the fixture, with nothing created under cwd.

(Found while implementing #272's sweep; the two prose workarounds above should be updated to link this issue when it's fixed.)

PR #392, opened 2026-07-29
fix(codegen): four toolchain-hygiene fixes, plus an unbounded manifest growth bug

### LINKED ISSUE #369: F12: no sweep for generator-owned files at retired default output paths — orphans stay, uncompiled and drift-tracked
## Summary

When a generated artifact's default output path moves between releases, the
file at the OLD path is never swept. It stays on disk: never compiled (no
`mod` hook, wrong directory), never removed, and still drift-compared — so
`check --regen-drift --write` keeps refreshing a file nothing builds. It
looks maintained and is dead.

## Evidence

Two instances of the same class in this repo:

1. Retired `--integration-output` default (`./src/integration_tests.rs` →
   `./programs/tests/integration_tests.rs`): three bundled examples carried
   an orphan at the old path for many releases. Removed by hand this week.

2. Retired `--test-output` default (`./src/tests.rs` →
   `./programs/tests/unit.rs`): the orphans are STILL in the tree.
   `examples/rust/{escrow,lending,multisig}/src/tests.rs` exist, carry the
   QEDGEN banner, and the sibling `src/lib.rs` has no `mod tests;`
   (`grep -c "mod tests" src/lib.rs` → 0), so nothing compiles them.

The only cleanup that exists inspects exactly one path — the CURRENT
`--integration-output` — and only when the target is not Quasar:

`crates/qedgen/src/run.rs:89-115` (`remove_obsolete_quasar_integration_file`),
called from `run.rs:2161` with `&integration_output`.

## Root cause

Cleanup is keyed to the current output path rather than to the set of paths
the generator has ever owned. Nothing records that a default moved.

## Proposed fix

A table consulted once per `codegen` run:

```rust
/// (retired path, what replaced it) — swept on every codegen run.
const RETIRED_OUTPUT_PATHS: &[(&str, &str)] = &[
    ("src/tests.rs",               "programs/tests/unit.rs"),
    ("programs/src/tests.rs",      "programs/tests/unit.rs"),
    ("src/integration_tests.rs",   "programs/tests/integration_tests.rs"),
    ("programs/src/integration_tests.rs", "programs/tests/integration_tests.rs"),
];
```

For each: if the file carries the QEDGEN banner, delete it and print one note
naming the replacement; if it looks user-edited, warn and keep it (the
existing function already makes that distinction correctly — reuse it).

This removes the class rather than the two known instances, and gives the
next default move a one-line landing spot.

## Gate

A CLI test that plants a banner file at a retired path, runs `codegen`, and
asserts the file is gone and the note was printed; plus a second case with a
non-banner file asserting it survives with a warning.

### LINKED ISSUE #370: F13: a relative --*-output that already contains the spec directory silently doubles it; --help does not state the resolution rule
## Summary

Relative `--*-output` paths resolve against the SPEC's directory, not the
invoker's cwd. That is deliberate and documented in
`references/cli/codegen.md`. The failure mode when a user types a
cwd-relative path anyway is a silently doubled directory, and the only signal
is the resolved path inside a success message.

## Repro

From a parent directory of the project:

```
$ qedgen codegen --spec proj/e.qedspec --target quasar --integration \
      --integration-output proj/tests/integration_tests.rs
  wrote proj/proj/tests/integration_tests.rs
$ find . -name integration_tests.rs
./proj/proj/tests/integration_tests.rs
```

Observed: exit 0, artifact written to `proj/proj/tests/`.
Expected: at minimum a warning that the given path repeats the spec directory.

## Root cause

`crates/qedgen/src/run_helpers.rs:56-62`:

```rust
pub(crate) fn resolve_output_against_spec(spec_dir: &Path, p: PathBuf) -> PathBuf {
    if p.is_absolute() { p } else { spec_dir.join(p).components().collect() }
}
```

`crates/qedgen/src/run.rs:1888-1912` applies it to all nine output flags. No
clap help text mentions the rule; only the reference doc does, so `--help`
does not carry it.

## Proposed fix

1. Warn on the doubling shape: when the output path is relative and its first
   component equals the last component of the spec directory, print

   > `--integration-output` resolves against the spec directory `<dir>/`;
   > `proj/tests/integration_tests.rs` resolved to
   > `proj/proj/tests/integration_tests.rs`. Pass a path relative to the spec
   > directory (`tests/integration_tests.rs`) or an absolute path.

   Two-line predicate, no behavior change, no false positive on a genuine
   nested directory of the same name as the project root (rare, and the
   warning is not fatal).

2. Add one sentence to the `--*-output` clap help text so `--help` states the
   resolution rule, instead of it living only in the reference doc.

## Gate

A CLI test asserting the warning fires for the doubling shape and does not
fire for a plain relative path.

### LINKED ISSUE #371: F14: the Parallax pin-liveness check is a manual RELEASING.md bullet, not a CI job — its failing condition breaks generated scaffolds for users

### LINKED ISSUE #390: F15: codegen --integration before the program scaffold writes a test whose dev-dependencies were never added, silently
## Summary

`qedgen codegen --integration` upserts the Parallax dev-dependency set into
the program crate's `Cargo.toml`. When that manifest does not exist yet, the
upsert has nothing to write into. It is skipped with no warning, and the
scaffold is written anyway.

The user is left with `programs/tests/integration_tests.rs` that cannot
resolve `parallax_svm`, and nothing said so. Generating the program crate
afterwards does not repair it: that pass writes a fresh `Cargo.toml` whose
only dev-dependency is `proptest`.

## Repro

```
$ mkdir -p proj/.qed && cd proj && git init -q
$ cp vault.qedspec .

# integration first, no program crate yet
$ qedgen codegen --spec vault.qedspec --target quasar --integration --output-dir programs
  wrote ./programs/tests/integration_tests.rs
$ ls programs/
tests

# now the program crate
$ qedgen codegen --spec vault.qedspec --target quasar --output-dir programs
Generated 7 files in programs
$ grep -c parallax-svm programs/Cargo.toml
0
$ sed -n "/dev-dependencies/,\$p" programs/Cargo.toml
[dev-dependencies]
proptest = "1"
```

Observed: exit 0 from both passes, an integration test on disk, and no
`parallax-svm` anywhere. `cargo test` fails with unresolved imports.

Expected: either the upsert runs (deferred until the manifest exists), or
`--integration` says it could not add the dev-dependencies and names the
command that would.

The contrast is the tell. In the working order, the second pass prints

```
  updated ./programs/Cargo.toml ([dev-dependencies] for Parallax)
  wrote ./programs/tests/integration_tests.rs
```

In the broken order that first line is simply absent, which is the only
signal a user gets and is not one anybody reads for.

## Root cause

The upsert is conditional on an existing manifest and has no else branch.
`--all` runs the program scaffold before the integration pass internally, so
the ordering that breaks is only reachable when a user passes `--integration`
on its own, which is the documented way to regenerate just that artifact.

## Proposed fix

1. When `--integration` finds no `Cargo.toml` at the output root, print a
   warning naming the fix (`run `qedgen codegen --spec … --target …` first,
   then re-run with `--integration``). Cheapest change, closes the silent
   part.
2. Better: have the program-scaffold pass preserve a `[dev-dependencies]`
   section it did not write, so the two passes commute. Today it rewrites
   the manifest wholesale, which is also why the order matters at all.

(2) is the real fix. Order dependence between two codegen selectors is
surprising on its own, warning or not.

## Gate

A CLI test running the two passes in the broken order and asserting
`parallax-svm` ends up in `[dev-dependencies]` (or, for fix 1, that the
warning fired). `parallax_integration_gate` deliberately runs the working
order and documents why, so it cannot catch this.

## How it was found

Writing the #383 gate. The gate has to run the program scaffold first, and
finding that out was trial and error rather than something the tool said.

abishekk92:
Closes #369. Closes #370. Closes #371. Closes #390.

Four small backlog items batched because each is a few lines and they touch
adjacent surfaces. A fifth bug fell out of testing the first one.

## #371 — the Parallax pin check runs itself now

`scripts/check-parallax-pin.sh` was wired only into the release checklist,
as a manual step. Its exit-1 condition is a user-facing break, not a release
concern: it fires when the pinned `parallax-svm` revision no longer exists
upstream, which means every generated integration scaffold stops resolving
its dev-dependencies, in the user's crate.

Now `.github/workflows/parallax-pin.yml` runs it weekly and
`scripts/release-gate.sh` calls it, so the release path is mechanical.
RELEASING.md 5a is reduced to a pointer, keeping only the genuinely manual
pin-bump procedure. Its own workflow rather than a job in `ci.yml`: adding
`schedule` there would put the whole matrix on a timer for one network
probe.

The script's own bump instructions were stale — they told you to update a
gate fixture manifest that #383 deleted. Corrected.

## #370 — the output-path doubling warning

Relative `--*-output` paths resolve against the SPEC's directory. That is
deliberate (#279), but a user standing one level up naturally writes
`proj/tests/x.rs`, which silently resolves to `proj/proj/tests/x.rs`. Both
exit 0.

Now:

```
$ qedgen codegen --spec proj/vault.qedspec --output-dir proj/programs
warning: `--output-dir` resolves against the spec directory `proj`, so `proj/programs` resolved to `proj/proj/programs`.
         Pass a path relative to the spec directory (`programs`) or an absolute path.
```

The predicate is narrow — relative AND first component equals the spec
directory's name — and split from the printing so it is testable. `--help`
now states the resolution rule too: it lived only in the reference doc.

## #369 — sweep retired output paths

`RETIRED_OUTPUT_PATHS` is consulted once per run. A generator-owned file
(QEDGEN banner) at a path a default has retired is deleted with a note
naming the replacement; a file without the banner is kept with a warning.
A retired path passed explicitly to THIS run is never swept, or codegen
would delete the artifact it was just told to write.

This removes the class rather than the instances, and the instances were
real: all four are gone from the bundled examples, including two the issue
listed as still present.

The multisig snapshot loses 1097 lines as a result. That is the orphan
`programs/src/tests.rs` leaving the staged tree — a file with no `mod` hook,
which nothing ever compiled. The snapshot had been pinning dead output as if
it were live, which is exactly the "looks maintained and is dead" shape the
issue describes.

## #390 — `--integration` before the program crate

The Parallax dev-dependency upsert needs a manifest to upsert into. With no
program crate yet there is none, so it was skipped silently and the scaffold
was written anyway, leaving a test that cannot resolve `parallax_svm` and no
diagnostic. It now says so and names the fix.

The issue also proposed making the two passes commute. Measured, they
already do in the other direction: `merge_cargo_toml` preserves a
`[dev-dependencies]` section it did not write, so bare `codegen` after
`--integration` keeps the Parallax pins. Verified rather than assumed — my
first report of this to the maintainer had the direction wrong.

## The fifth bug: the manifest grew on every run

Found while regenerating the examples for #369. `push_section` appended a
separator to a body that already carried one, so every `codegen` run added
one blank line per section, unbounded:

```
run 1: 23 lines, 6 blank
run 2: 24 lines, 7 blank     <- before the fix, kept climbing:
run 3: 32 lines, 15 blank
run 4: 34 lines, 17 blank
run 5: 36 lines, 19 blank
```

Nothing broke, which is why it went unnoticed. The file just grew and never
matched its committed copy again.

Fixing the growth exposed a second half: the merged form carried one
trailing blank line the greenfield render did not. `check --regen-drift`
regenerates into an empty temp directory — the greenfield path — and
compares against a committed file that has only been through the merge path,
so that single line would have read as drift on every example, forever. Both
forms are now byte-identical, asserted by a test.

## Verification

- `check --regen-drift`: clean, 7 examples.
- `parallax_integration_gate` (`--ignored`): passes.
- Full `cargo test`: 40 targets green. Clippy and rustfmt clean.
- `check-readme-drift.sh`: no drift, 22 commands documented.
- Snapshots refreshed; every non-multisig change is the blank-line
  normalization, and the multisig change is the orphan described above.
- The #370 warning and the #369 sweep were both exercised end to end against
  the real binary, including the negative cases (plain relative path stays
  quiet; a non-banner file at a retired path survives with a warning).


abishekk92:
reopening to re-fire the pull_request event; no Actions run was created for this branch

abishekk92:
Superseded — no Actions run was ever created for this PR (Greptile only). Reopening as a fresh PR from the same branch to get CI.

